PR #6, opened 2015-09-04
Make the tenant connection configurable

thijsvdanker:
I would like to make the tenant connection for TenantModels configurable so I can change the connection when running tests by changing the environment properties.

At the current moment I have to set-up an entire multi-tenant environment for each functionality that uses a tenant model. While that could be exactly what you need (testing multi-tenancy), at other times you only want to test a single slice of functionality.

__Use case:__
I want to test if my API-endpoint is formatting my model data properly.
This is not related to my multi-tenancy setup.

__Usage:__
By default the TENANT_CONNECTION resolves to tenant.
By setting TENANT_CONNECTION in your phpunit.xml file to `<env name="TENANT_CONNECTION" value="mysql"/>` the models are resolved from your default mysql database.
-Or-
By changing it on a per test basis:
```php
Class APIResponseTest extends TestCase {

    /**
     * Setup the test environment.
     *
     * @return void
     */
    public function setUp()
    {
        putenv('TENANT_CONNECTION=mysql');
        parent::setUp();
    }
```

Let me know if it makes sense to add this to the package, otherwise people can just create their own implementation of TenantModel and use these snippets to simplify some test set-ups.

### DIFF
--- a/src/Abstracts/Models/TenantModel.php
+++ b/src/Abstracts/Models/TenantModel.php
@@ -6,5 +6,11 @@ use HynMe\Framework\Models\AbstractModel;
 
 class TenantModel extends AbstractModel
 {
-    protected $connection = 'tenant';
+    protected $connection;
+
+    public function __construct(array $attributes = [])
+    {
+        $this->connection = env('TENANT_CONNECTION', 'tenant');
+        parent::__construct($attributes);
+    }
 }


thijsvdanker:
I've expanded the setup in a trait that also runs my tenant-migrations against the database.
Might be useful for someone else:
https://gist.github.com/thijsvdanker/ec1ef6defe34d903f089

thijsvdanker:
Tests didn't fail because of the PR but because composer had issues.

Luceos:
Yes the tests aren't attuned for PR's yet.

I'm not sure what you are proposing and will get in touch about this asap, haven't seen this PR popup on my radar. Did you also check: https://github.com/laraflock/multi-tenant/blob/master/src/Tenant/DatabaseConnection.php for possible solutions to set a connection as current.

thijsvdanker:
Hrmm so stubbing the getCurrent method on the DatabaseConnection class would probably work as well right?

Luceos:
I'm thinking of making the tenant connection configurable anyway. Now it simply generates a db configuration dynamically and uses that to connect. 

If I'm not mistaking your question you'd like to set any/all tenants to use a preconfigured database connection. So in order to get there, there should be some kind of setter, like `$tenant->usesConnection($name);` that will force a tenant site to use a different connection.

Would this be a solution?

Luceos:
Committed 83a105510530e1a859a2da50829fb6ae3aac4fde which should give you proper opportunity to play with.

PR #92, opened 2016-06-08
Make Insight analysis happy

bkintanar:
Also changed DOS line endings of `views/webserver/nginx/configuration.blade.php` to Unix.


luceos:
It would be better if you separated the fixes from the refactoring. Now I won't be able to merge because you refactored image_cache and base_old, but none of the others. As such it's only a partial solution. If you would be able to PR the singleton fix, I'll merge immediately 👍

PR #120, opened 2016-07-29
Change --tenant to --customer

bkintanar:


luceos:
Thanks I missed that one ;)

PR #129, opened 2016-08-11
Fix in down() function.

### LINKED ISSUE #128: Bug In Migration
The down() function for RenameTenantsToCustomers2 has some bugs.

bkintanar:
Fixes #128


luceos:
Fixes #128

PR #136, opened 2016-08-17
Initial Support for Laravel 5.3

luceos:
This is a WIP. Awaiting release of 5.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - mysql -e "create database hyn;" -uroot
   - mysql -e "grant all privileges on *.* to 'travis'@'localhost' with grant option;" -uroot
   # create a new laravel project in the subfolder laravel (default composer behaviour)
-  - composer require laravel/laravel 5.2
+  - composer require laravel/laravel dev-develop
   # load pheanstalk package to talk with beanstalkd queue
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ install:
   - mysql -e "create database hyn;" -uroot
   - mysql -e "grant all privileges on *.* to 'travis'@'localhost' with grant option;" -uroot
   # create a new laravel project in the subfolder laravel (default composer behaviour)
-  - composer require laravel/laravel 5.2
+  - composer require laravel/laravel dev-develop
```
luceos, line 32: We need to test against both 5.2 and 5.3.

composer.json
```diff
@@ -16,10 +16,11 @@
     },
     "require": {
         "php": ">=7.0.0",
-        "laravel/framework": "~5.2.0",
+        "laravel/framework": "~5.3.0",
```
luceos, line 19: We'll not hardcode 5.3, because then we won't support 5.2

.travis.yml
bkintanar, line 32: Ok. I'll work on this later.

luceos:
PR was a bit premature.

Issue #152, opened 2016-09-14
How to overwrite the default views files through the multi-tenant folder?


deleugpn:
I'm trying to build a Hello World project with 2 customers and see the power of the package for myself, but I'm stuck at the **resources** and **routes** elements now.
## The problem

When accessing my different hostnames (`home.php7.local` and `customer-1.php7.local`), I see `welcome.blade.php` that is located under `hello-world/resources/views/welcome.blade.php`. I tried creating different files under their own storage folder, but no luck so far.
I have one file at `/hello-world/storage/multi-tenant/1-home-php7-/views/welcome.blade.php` and a different one at `/hello-world/storage/multi-tenant/2-customer-1/views/welcome.blade.php`. How do I get them to overwrite the default `welcome.blade.php` located in the `resources/views` folder?

For more information, the next session gives a little background on how I got here.
## Background
- I installed a brand new Cent OS 7 VM from scratch using the minimal iso and Virtual Box (No Vagrant), installed **PHP 7.0**, **MySQL 5.7** and set a shared folder to point to `/var/www/html/`.
- After that, I created a clean Laravel project and installed `hyn/multi-tenant: 2.0.x@dev` through composer.
- I made sure to [Integrate Into Laravel](https://hyn.readme.io/docs/integrate-into-laravel) and [Configure](https://hyn.readme.io/docs/configuration) it following the docs.
- **Note:** _I didn't run the queue at all because I would like to configure Apache myself because when the time for deployment comes, I will be using a cPanel environment. Although I have root access to it, I can't allow third-party libraries to change cPanel Apache configuration files._
## Tenancies

I now have 2 **customers**, 2 **hostnames** and 2 **websites**. One is `home.php7.local` and the other is `customer-1.php7.local`. The following snippet represents the Apache Configuration for **customer-1**, but **home** is pretty similar, changing just the _alias_ value and the _logs_.

```
<VirtualHost *:80>
  ServerName customer-1.php7.local
  DocumentRoot "/var/www/html/hello-world/public/"
  alias "/media/" "/var/www/html/hello-world/storage/multi-tenant/2-customer-1/media/"
  <Directory "/var/www/html/hello-world/public/">
    Options Indexes FollowSymlinks MultiViews
    AllowOverride All
    Require all granted
  </Directory>
  ErrorLog "/var/log/httpd/error-customer-1-hello-world.log"
  ServerSignature Off
  CustomLog "/var/log/httpd/access-home-customer-1-world.log" combined
[...]

PR #159, opened 2016-09-23
Change loadViews to prepend tenant location so customized views will have priority

deleugpn:
As mentioned on issue #152, it seems like a good idea to assume whenever a tenant-view exists, it should override the default system-view (otherwise why make the tenant view at all?). 

The change here is to simply `prepend` the tenant-view location path instead of just appending it. That way, when finding a view file, Laravel Finder will automatically take the tenant-view first, thus allowing the tenant to have precedence.

I was not able to come up with a Test Code for this small change, unfortunately.

### DIFF
--- a/src/Tenancy/Tenant/Directory.php
+++ b/src/Tenancy/Tenant/Directory.php
@@ -214,7 +214,7 @@ class Directory implements DirectoryContract
     protected function loadViews($app)
     {
         if (!$this->disallowed('views') && $this->views() && File::isDirectory($this->views())) {
-            $app['view']->addLocation($this->views());
+            $app['view']->getFinder()->prependLocation($this->views());
         }
     }
 


luceos (changes requested):
What would be better is to get a new config parameter into the multi-tenant configuration file that allows anyone to append or prepend. As making prepend default, I think we should keep append default for now. What do you think @Deleugpn ?

deleugpn:
To be honest, I see no reason why someone would have a tenant-specific view page and want it to default to the system-wide one. It makes the tenant-specific view useless. That said, I also don't think a config flag would be of harm in any way, so I'm going to try and follow that reasoning.

luceos:
Makes sense. If you can make the tenant overrule by default, it would be fine. Anyway I want things to be as flexible as possible.

luceos (approved):

luceos:
@Deleugpn I have reconsidered your comment and it makes sense to stick to your logic for now. Thank you.

deleugpn:
I'd like to apologize, the last couple of weeks have been crazy and I haven't cleared any spare time, so that's why I didn't work towards your suggestion for this.
Regardless, thanks for considering it, I believe it will be a good thing for the package.

luceos:
@Deleugpn no need to apologize, this is how life and open source works :+1:

Issue #104, opened 2016-06-11
New migration doesn't run on tenant


bkintanar:
Ok here's the gist of my situation. I created a set of new migrations and upon issuing `php artisan migrate --tenant=all` it doesn't migrate the tenant database with the new migrations. Please see steps below.

PS. This only works if you apply my PR: https://github.com/hyn/multi-tenant/pull/103
1. blank master database
2. no tenant databases yet
3. I have migrations after hyn/multi-tenant migrations.
4. I then run `multi-tenant:setup` as seen below to create new tenant. New migrations as highlighted in the screenshot shows at the bottom. No surprises there since that's how it should work.
   <img width="1158" alt="screen shot 2016-06-11 at 9 36 42 am" src="https://cloud.githubusercontent.com/assets/685928/15982455/8e77da84-2fba-11e6-957d-f23f30ff6bb3.png">
5. I then ran migrations for tenant `php artisan migrate --tenant=all`
6. As you can see on the screenshot the 3 migrations from above doesn't appear in the tenant migration.
   <img width="570" alt="screen shot 2016-06-11 at 9 36 59 am" src="https://cloud.githubusercontent.com/assets/685928/15982456/8e7a5e26-2fba-11e6-930e-8d164ceacb9e.png">
7. I reset the master database, deleted all the tables, removed the newly created tenant.
8. I then tested it out and renamed all 3 migrations in such a way that it will show up first on the list and that the hyn/multi-tenant migrations will be shown last. As you can see on the screenshot it works as expected on the master database, the 3 migrations are now shown first.
   <img width="1154" alt="screen shot 2016-06-11 at 9 42 40 am" src="https://cloud.githubusercontent.com/assets/685928/15982457/8e7edf0a-2fba-11e6-913f-ad0f30ddc941.png">
9. I then tried again and run the migrations on the tenant created. This time the 3 migrations hows on top of the hyn/multi-tenant migrations as expected.
   <img width="545" alt="screen shot 2016-06-11 at 9 43 08 am" src="https://cloud.githubusercontent.com/assets/685928/15982458/8e8752f2-2fba-11e6-9b4f-6a7103fbaae6.png">

luceos:
Didn't I do the same here?

https://github.com/hyn/multi-tenant/pull/98/files

bkintanar:
@Luceos I'm afraid not. There are two hostnames_table around. hmt_hostnames ad hws_ssl_hostnames and you missed the following files below.

```
[...]

PR #162, opened 2016-10-13
Migration fixes past 2014

psaunders:


luceos:
Hello @psaunders can you explain what this PR is trying to fix? Thanks!

psaunders:
The tenant migrations are running the migration scripts on the HYN database. A few migrations like HmtTenantsTable run fine because the have an existence check like:

```
!Schema::connection(DatabaseConnection::systemConnectionName())->hasTable('tenants')) 
```

Later migrations (after 2014) don't have these checks so the migration causes a duplicate table error and don't progress for the tenant tables. I would look into having a special folder for just the HYN migrations so they're not mixed in with the tenant migrations, but for now this gets it working.

psaunders:
Probably related to this one: https://github.com/hyn/multi-tenant/issues/104

PR #271, opened 2017-10-08
Seeds after creating website

Thijmen:
- Added option to seed a created website after creating.
- Added option to the tenancy:db:seed to seed an individual website by id.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -262,7 +262,6 @@ class Connection
         $code = $this->artisan->call('tenancy:db:seed', $options);
         $this->purge($this->tenantName());
         return $code == 0;
-
     }
 
 
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -106,15 +106,8 @@ class Connection
         return $this->db->connection($this->tenantName());
     }
 
-    /**
-     * @param Hostname|Website $to
-     * @param null $connection
-     * @return bool
-     */
-    public function set($to, $connection = null): bool
+    private function convertWebsiteOrHostnameToWebsite($to)
     {
-        $connection = $connection ?? $this->tenantName();
-
         $website = null;
 
         if ($to instanceof Hostname) {
@@ -124,6 +117,20 @@ class Connection
         if ($to instanceof Website) {
             $website = $to;
         }
+
+        return $website;
+    }
+
+    /**
+     * @param Hostname|Website $to
+     * @param null $connection
+     * @return bool
+     */
+    public function set($to, $connection = null): bool
+    {
+        $connection = $connection ?? $this->tenantName();
+
+        $website = $this->convertWebsiteOrHostnameToWebsite($to);
 
         if ($website) {
             // Sets current connection settings.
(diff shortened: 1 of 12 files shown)


src/Database/Connection.php
```diff
@@ -236,6 +236,35 @@ public function migrate($for, string $path = null)
         return $code === 0;
     }
 
+    public function seed($for)
+    {
+        $this->set($for, $this->tenantName());
```
luceos: Use `migrationName` not `tenantName`. MigrationName is intended for short lived connections, eg during mass changes and commands.

src/Database/Connection.php
```diff
@@ -236,6 +236,35 @@ public function migrate($for, string $path = null)
...
+        $website = null;
+
+        if ($for instanceof Hostname) {
+            $website = $for->website;
+        }
+
+        if ($for instanceof Website) {
+            $website = $for;
+        }
+
```
luceos: This logic is repeated from the migrate method. You should make this [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Separate both pieces of code into a new private or protected method.

src/Database/Connection.php
```diff
@@ -236,6 +236,35 @@ public function migrate($for, string $path = null)
...
+
+
+        $options = [
+            '--database' => $this->tenantName(),
+            '--websiteid' => $website->id,
+            '--class' => config('tenancy.db.tenant-seed-after-created-website')
```
luceos: The seed method on this class should be generic; you should be able to call them with any other path/class. The seed method therefor should hold another argument `string $class` which you pass the config value from the outside. Hardcoding this config value inside this class is not the way to go.

src/Listeners/Database/MigratesTenants.php
```diff
@@ -46,8 +48,19 @@ public function subscribe(Dispatcher $events)
...
+            $resultUsed = true;
+        }
+
+        if (!$result && $resultUsed) {
+            return $result;
+        }
+
+        if (config('tenancy.db.tenant-seed-after-created-website')) {
+            $this->emitEvent(new Events\Websites\Migrated($event->website));
         }
```
luceos: There are several things I would like to see differently in this class. The package has to be flexible, the way you implemented it you are assuming that you only want to trigger a Migrated event in case someone configures seeds to work. In fact this implementation shows how much you focussed on your particular use case, which I totally get btw. 

So what would I do;

- remove the unnecessary temporary variables.
- emit the event if the command succeeded

src/Listeners/Database/SeedsTenants.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     *
+     * @param WebsiteEvent $event
+     * @return bool
+     */
+    public function seed(WebsiteEvent $event) : bool
+    {
+        return $this->connection->seed($event->website);
```
luceos: I would check here whether the configuration key is set, just like I did [here](https://github.com/codemotionmedia/multi-tenant/blob/bcb99af1a36fdec421ad9353731c40598e680faf/src/Listeners/Database/MigratesTenants.php#L53) and then run the seed.

src/Repositories/WebsiteRepository.php
```diff
@@ -63,6 +63,15 @@ public function findByUuid(string $uuid)
     }
 
     /**
+     * @param string $id
+     * @return Website|null
+     */
+    public function findById(string $id)
+    {
+        return $this->website->newQuery()->where('id', $id)->first();
```
luceos: Use `$this->query()` instead of `$this->website->newQuery()`.

luceos (changes requested):
I'm very glad to see you've made such a great effort of this PR. It's really looking great. I've made some generic comments.

There are a few times in your PR I could clearly see that you've written the code to match your particular use case. The package aims to be flexible for a broad audience though. Eg, being able to disable migrations for new tenants, but still seed.

Also I'm not too sure the implementation of the SeedCommand is that great. Aside from PSR2 standards not being applied it also undoes some of the code I've written with specific reasoning (eg using a migration connection).

If you're willing to do some further improvements, I'll take a look afterwards and patch up the last pieces so it is included.

Thank you again! :hugs:

PR #280, opened 2017-10-12
Improved validation of hostnames

Thijmen:
It should be possible to update a hostname and set `redirect_to`, `customer_id` and `under_maintenance_since` to `null`.

### DIFF
--- a/src/Validators/HostnameValidator.php
+++ b/src/Validators/HostnameValidator.php
@@ -23,7 +23,7 @@ class HostnameValidator extends Validator
         'redirect_to' => ['nullable', 'string', 'url'],
         'force_https' => ['boolean'],
         'under_maintenance_since' => ['nullable', 'date'],
-        'website_id' => ['integer', 'nullable', 'exists:%system%.websites,id'],
+        'website_id' => ['nullable', 'integer', 'exists:%system%.websites,id'],
         'customer_id' => ['nullable', 'integer', 'exists:%system%.customers,id'],
     ];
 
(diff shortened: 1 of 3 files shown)


src/Validators/HostnameValidator.php
```diff
@@ -20,20 +20,20 @@ class HostnameValidator extends Validator
 {
     protected $create = [
         'fqdn' => ['required', 'string', 'unique:%system%.hostnames,fqdn'],
-        'redirect_to' => ['string', 'url'],
+        'redirect_to' => ['nullable','string', 'url'],
         'force_https' => ['boolean'],
-        'under_maintenance_since' => ['date'],
+        'under_maintenance_since' => ['nullable', 'date'],
         'website_id' => ['integer', 'exists:%system%.websites,id'],
```
luceos: website_id should also be nullable

src/Validators/HostnameValidator.php
```diff
@@ -20,20 +20,20 @@ class HostnameValidator extends Validator
...
 
     protected $update = [
         'id' => ['required', 'integer'],
         'fqdn' => ['required', 'string', 'unique:%system%.hostnames,fqdn'],
-        'redirect_to' => ['string', 'url'],
+        'redirect_to' => ['nullable','string', 'url'],
         'force_https' => ['boolean'],
-        'under_maintenance_since' => ['date'],
+        'under_maintenance_since' => ['nullable', 'date'],
         'website_id' => ['integer', 'exists:%system%.websites,id'],
```
luceos: website_id should also be nullable

luceos (changes requested):
Thank you very much, great addition. Nice to see some additional tests. Just one comment about website_id being nullable too 😊

luceos (approved):

PR #281, opened 2017-10-13
Fixes Maintenance-mode on the middleware

Thijmen:
This should fix the middleware.
As CurrentHostname is injected, it should be nullable. It could be applied to a route for a website which isn't identified. 

Added also some tests, plus, I made the `under_maintenance_since` a date in the `$dates` object.

### DIFF
--- a/tests/unit-tests/EnvironmentTest.php
+++ b/tests/unit-tests/EnvironmentTest.php
@@ -58,7 +58,8 @@ class EnvironmentTest extends Test
         $this->assertNull($identified);
 
         $this->hostname->save();
-        $this->app->make('config')->set('tenancy.hostname.default', $this->hostname->fqdn);
+        
+        config(['tenancy.hostname.default' => $this->hostname->fqdn]);
 
         $identified = $this->app->make(CurrentHostname::class);
 
@@ -73,6 +74,7 @@ class EnvironmentTest extends Test
     public function we_can_set_current_hostname_to_null_on_hostname_action_middleware()
     {
         $middleware = new HostnameActions(null, app()->make(Redirector::class));
+        
         $this->assertNotNull($middleware);
     }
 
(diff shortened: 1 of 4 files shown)


luceos (changes requested):
You're putting too many assertions inside the same test method. Please split them up and assign method names that indicate their purpose.

tests/unit-tests/EnvironmentTest.php
```diff
@@ -62,6 +66,48 @@ public function identifies_hostname()
...
+    public function maintenance_test()
+    {
+        try {
+            $middleware = new HostnameActions(null, app()->make(Redirector::class));
+        } catch (\TypeError $ex) {
+            $this->fail('It should accept null as CurrentHostname, as this middleware should be able to run on non-identified websites.');
+        }
+
+
+        $this->hostname->save();
```
luceos: Use config function to set the value. Much cleaner.

tests/unit-tests/EnvironmentTest.php
```diff
@@ -62,6 +66,48 @@ public function identifies_hostname()
...
+     * @test
+     */
+    public function maintenance_test()
+    {
+        try {
+            $middleware = new HostnameActions(null, app()->make(Redirector::class));
+        } catch (\TypeError $ex) {
+            $this->fail('It should accept null as CurrentHostname, as this middleware should be able to run on non-identified websites.');
+        }
+
```
luceos: One empty line too many.

tests/unit-tests/EnvironmentTest.php
```diff
@@ -62,6 +66,48 @@ public function identifies_hostname()
         $this->assertEquals($identified->fqdn, $this->environment->hostname()->fqdn);
     }
 
+    /**
+     * @test
+     */
+    public function maintenance_test()
+    {
+        try {
+            $middleware = new HostnameActions(null, app()->make(Redirector::class));
```
luceos: Import all classes, also those in the root namespace

tests/unit-tests/EnvironmentTest.php
```diff
@@ -62,6 +66,48 @@ public function identifies_hostname()
...
+        } catch (\TypeError $ex) {
```
luceos: Split this test method up and use phpdoc expectedException to anticipate an exception

luceos (approved):

luceos:
Thank you 🤗 . Will bring this along in the next minor!

PR #294, opened 2017-10-19
Tenant specific migrations

wertmenschen:
With this commit tenants can have migrations that are run only for them. If a tenant migration path is configured and there is a subdirectory with the website UUID in this path, the contained migrations will be run.

### DIFF
--- a/src/Traits/MutatesMigrationCommands.php
+++ b/src/Traits/MutatesMigrationCommands.php
@@ -53,6 +53,7 @@ trait MutatesMigrationCommands
 
         $this->processHandle(function ($website) {
             $this->connection->set($website, $this->connection->migrationName());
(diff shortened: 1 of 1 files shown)


luceos:
I think the idea behind your PR is great. However I am afraid by implementing it this way it might affect migrations in unforeseen situations. By modifying the trait you would always look for a subdirectory matching the uuid of the website, no matter why, when or how you run the migration command.

In addition setting a global website property isn't really a pretty way of implementing this.

I would consider such a feature if;

- you could set a flag in the configuration to look for deeper directories based on the uuid
- the path would be appended differently; I am not sure how yet. Perhaps if the flag is set it would be added during the process of auto migrating a new tenant?

luceos:
No response.

PR #298, opened 2017-10-30
Customisable Models

TNovalis:
Changelog:
* Added the ability to customize models in config
* Added config merging so updating your config is not *explicitly* required
* Added migration publishing, so changing the default migrations is simple
* Fixed some typos in events and config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /vendor
 composer.phar
 composer.lock
+/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /vendor
 composer.phar
 composer.lock
-/.idea
--- a/src/Contracts/CurrentHostname.php
+++ b/src/Contracts/CurrentHostname.php
@@ -33,13 +33,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsTo;
  */
 interface CurrentHostname
 {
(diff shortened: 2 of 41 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 /vendor
 composer.phar
 composer.lock
+/.idea
```
luceos: IDE specific folders should be excluded using a global gitignore file, see:

https://help.github.com/articles/ignoring-files/

Otherwise you'd have to take into consideration every single IDE any contributor ever made.

src/Contracts/CurrentHostname.php
```diff
@@ -14,6 +14,32 @@
...
+ * @property Carbon $deleted_at
+ * @property int $website_id
+ * @property Website $website
+ * @property int $customer_id
+ * @property Customer $customer
+ */
 interface CurrentHostname
 {
+    /**
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
```
luceos: You've already imported these, why not simplify the class here as well?

luceos (changes requested):
Wow, I'm impressed!

Just a few things I was thinking of, otherwise this is AOK for a merge into master.

TNovalis:
Removed the fluff

PR #313, opened 2017-11-02
fixes #306, by using models in seeder these would fail

### LINKED ISSUE #306: Seeding for a specific Website is not working
__Description__

If I try to execute a Seeder manually with:
```bash
php artisan tenancy:db:seed --class=...Seeder --website_id=5
```
the following error shows up
```
Database [tenant] not configured.
```
shows up
---

__Actual behavior__

The seeder tries to use the configured tenant connection

__Expected behavior__

The seeder should use the connection from the website given by the website_id

---
__Information__

- hyn/multi-tenant version: 5.0.8
- laravel version: 5.5.19
- database driver and version: mysql 5.7
- webserver software and version: apache 2.4
- php version: 7.1

luceos:


src/Database/Connection.php
```diff
@@ -25,15 +25,13 @@
...
 use Hyn\Tenancy\Events;
-use InvalidArgumentException;
 
 class Connection
 {
     use DispatchesEvents;
 
     const DEFAULT_SYSTEM_NAME = 'system';
     const DEFAULT_TENANT_NAME = 'tenant';
-    const DEFAULT_MIGRATION_NAME = 'tenant-migration';
```
Thijmen, line 36: Wouldn't it be better to deprecate this const in stead of removing it? Removing it would break stuff for users.

luceos: I should have done that indeed. Rushed decision.

Thijmen: Can we still revert this in a new release?
EDIT: Great job, thanks for updating!

PR #315, opened 2017-11-04
Forcing artisan calls.

Thijmen:
When on production, this is needed, otherwise artisan will prompt an "Are you sure you want to run this on production" message.

I stumbled upon this when running it in production because of migrations not being run.


luceos:
Asking to continue is default behavior in laravel. I don't see why this package should work differently. Using force should be the normal way to ignore the confirmation.

Issue #322, opened 2017-11-15
Backup / restore of tenants


ghost:
__Description__

In case of an issue with your server, the system database and tenant db's will be backupped. After setting up a new server, you will pull the repository and reinstall the application. Run composer update and your NPM packages, if any. 

---

__Actual behavior__

This results in the behavior of a fresh installation and your databases need to be recreated from scratch too. Since you don't know the username and password? One way to recreate them, is the use of > vendor/hyn/multi-tenant/src/Providers/Tenants/PasswordProvider.php which will create the password. In your system db, you have the UUID's of your websites and this should do the trick? Thanks, @bkintanar for thinking with me!

__Expected behavior__

This needs a bit of documentation because at the time of emergency you don't want to spend this on finding this solution ;)

---
__Information__

- hyn/multi-tenant version: all
- laravel version: all
- database driver and version: all
- webserver software and version: all
- php version: all

---

__tenancy.php config__

```php

```

__webserver.php config__

```php

```

---

__Error log__

```text

```

Thijmen:
I agree, it should be documented. We have it on our roadmap and we will notify you on this issue when we added the documentation. If you want to help, feel free to create a pull request @ https://github.com/hyn/laravel-tenancy.com

bkintanar:
For the past two days, @Develeapment and I created a solution. As long as you have the system db intact, it can regenerate the tenant dbs with the command we created.

PR #326, opened 2017-11-17
Add restore command

bkintanar:
Based on #322

### DIFF
--- a/tests/unit-tests/Commands/RecreateCommandTest.php
+++ b/tests/unit-tests/Commands/RecreateCommandTest.php
@@ -14,14 +14,9 @@
 
 namespace Hyn\Tenancy\Tests\Commands;
 
-use Hyn\Tenancy\Traits\DispatchesEvents;
-use Hyn\Tenancy\Events\Websites;
-use Hyn\Tenancy\Models\Website;
-use Illuminate\Support\Facades\DB;
-use Illuminate\Database\QueryException;
 use App\Console\Kernel;
+use Hyn\Tenancy\Traits\DispatchesEvents;
 use Illuminate\Contracts\Foundation\Application;
-use Doctrine\DBAL\Driver\PDOException;
 
 class RecreateCommandTest extends DatabaseCommandTest
 {
@@ -58,14 +53,14 @@ class RecreateCommandTest extends DatabaseCommandTest
         $this->website->save();
 
         try {
-            if ($this->connection->get()->getSchemaBuilder()->hasTable('migrations')) {
-                return "ok";
+            if (!$this->assertFalse($this->connection->get()->getSchemaBuilder()->hasTable('migrations'))) {
+                $this->fail('`migrations` table in tenant db still exists.');
             }
-
-            $this->fail('Command didn\'t fire exception');
-        } catch (PDOException $e) {
-            $this->artisan->call('tenancy:recreate');
+        } catch (\Exception $e) {
+            // Surpress exception
         }
+
+        $this->artisan->call('tenancy:recreate');
         
         $this->connection->set($this->website);
 
--- a/tests/unit-tests/Commands/RecreateCommandTest.php
+++ b/tests/unit-tests/Commands/RecreateCommandTest.php
@@ -48,8 +48,6 @@ class RecreateCommandTest extends DatabaseCommandTest
         $this->connection->set($this->website);
 
         $this->assertTrue($this->connection->get()->getSchemaBuilder()->hasTable('migrations'));
-
-        $this->connection->purge();
         
         $this->websites->delete($this->website);
 
--- a/tests/unit-tests/Commands/RecreateCommandTest.php
+++ b/tests/unit-tests/Commands/RecreateCommandTest.php
@@ -49,6 +49,8 @@ class RecreateCommandTest extends DatabaseCommandTest
 
         $this->assertTrue($this->connection->get()->getSchemaBuilder()->hasTable('migrations'));
 
+        $this->connection->purge();
+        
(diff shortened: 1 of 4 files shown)


src/Commands/RestoreCommand.php
```diff
@@ -0,0 +1,65 @@
...
+    protected $description = 'Restore all tenant databases present in the system db.';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
```
luceos: unnecessary construct, drop the construct method all together so it falls through to the parent without calling it explicitly

luceos (changes requested):
I like the idea, do we need to add any checks at all before trying to recreate tenants? @Thijmen ?

src/Commands/RestoreCommand.php
bkintanar: hmmm. this is an output of make:command. I'll remove it.

bkintanar:
it will throw an error if you run this command and there are databases that exists.

src/Commands/RestoreCommand.php
luceos: Fix the above and I think we can merge it as is. Make sure to remove the TODO comments too for now, no need to keep stuff like that in versioned code.

And .. make sure to ask confirmation!

src/Commands/RestoreCommand.php
```diff
@@ -0,0 +1,53 @@
...
+class RestoreCommand extends Command
+{
+    use DispatchesEvents;
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'tenancy:restore';
```
Thijmen: Not quite sure about the naming.. it's not restoring, but recreating the databases. If I see "restore", I am thinking of restoring a backup or something, not just recreating the database.

src/Commands/RestoreCommand.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Restore all tenant databases present in the system db.';
```
Thijmen: Same as above, I think restore isn't a good name for this purpose.

src/Commands/RestoreCommand.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Hyn\Tenancy\Commands;
+
+use Hyn\Tenancy\Models\Website;
+use Illuminate\Console\Command;
+use Hyn\Tenancy\Traits\DispatchesEvents;
+use Hyn\Tenancy\Events\Websites as Events;
+
+class RestoreCommand extends Command
```
Thijmen: Command name should be different; see below.

src/Commands/RestoreCommand.php
```diff
@@ -0,0 +1,53 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $websites = Website::all();
+
+        foreach ($websites as $website) {
+            $this->emitEvent(new Events\Created($website));
```
Thijmen: I think we need to do a few things here;
- Check if the database already exists
- If not; create. If it does exists, we can get an exception here. 

Another option would be a try-catch and in the catch a message in the output that it failed to create the database.  Otherwise, the command would fail here and other websites won't be created as a new database.

Thijmen (commented):
I have given my thoughts about the code.

luceos:
@bkintanar are you up to add that database exists check? You could use the System Connection instance for that.

luceos (dismissed):

luceos (changes requested):
One last thing @bkintanar : a test that confirms that after deleting a tenant database and other things, we can recreate it.

tests/unit-tests/Commands/RecreateCommandTest.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->connection->migrate($this->website, __DIR__ . '/../../migrations');
+
+        $this->connection->set($this->website);
+
+        $this->assertTrue($this->connection->get()->getSchemaBuilder()->hasTable('migrations'));
+
+        // emit delete event to delete database.
+        $this->emitEvent(new Websites\Deleted($this->website));
+
+        if (!$this->connection->get()->getSchemaBuilder()->hasTable('migrations')) {
```
Thijmen: You need to assert here that the website has been deleted. After that, you can call `tenancy:recreate`.

ashokgelal:
No love for MySQL/ MariaDB?

Thijmen:
It should also support MySQL/MariaDB @ashokgelal, why do you think we don't show them love?

luceos:
@ashokgelal our tests confirm the workings of this command against three db drivers: mariadb, mysql and pgsql. In fact we had to patch up the PR to support pgsql completely, as mysql and mariadb already worked out of the box.

bkintanar:
@luceos maybe just to be on the safe side, we need to add that patch to MariaDB and MySQL as well?

ashokgelal:
Because I saw changes to https://github.com/hyn/multi-tenant/pull/326/files#diff-432e0887e3650de29d527cbcd5de53da but not to the MariaDB counterpart. The patch is also important for consistency as now with these changes a tenant user gets deleted when deleting a pgsql customer/website but not when deleting a mysql customer/website.

luceos:
@ashokgelal good point, just to prevent cluttering the database @bkintanar you're right then too.

PR #333, opened 2017-11-21
fixes #310 support for queue handling

### LINKED ISSUE #310: Make queue handling better for tenants
Current there is no easy way to sent jobs meant for tenants into the queue. I recommend sending along the tenant (website) id and loading that one again, activating the tenant if necessary to resolve the database connection.

Queue handling should be improved, I was thinking of a Trait or Abstract Job, but this has to be specced out better.

luceos:
Adds a trait that replaces the SerializesModel trait to serialize/unserialize attributes of a job, injecting the currently active tenant on __sleep and re-activating it on __wakeup.

### DIFF
--- a/src/Traits/ConvertsEntityToWebsite.php
+++ b/src/Traits/ConvertsEntityToWebsite.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Hyn\Tenancy\Traits;
+
+use Hyn\Tenancy\Contracts\Hostname;
+use Hyn\Tenancy\Contracts\Website;
+
+trait ConvertsEntityToWebsite
+{
+    /**
+     * @param $to
+     * @return Hostname|Website|null
+     */
+    protected function convertWebsiteOrHostnameToWebsite($to)
+    {
+        $website = null;
+
+        if ($to instanceof Hostname) {
+            $website = $to->website;
+        }
+
+        if ($to instanceof Website) {
+            $website = $to;
+        }
+
+        return $website;
+    }
+}
--- a/src/Traits/ConvertsEntityToWebsite.php
+++ b/src/Traits/ConvertsEntityToWebsite.php
@@ -21,7 +21,7 @@ trait ConvertsEntityToWebsite
 {
     /**
      * @param $to
-     * @return Hostname|Website|null
+     * @return Website|null
      */
     protected function convertWebsiteOrHostnameToWebsite($to)
     {
--- a/src/Traits/ConvertsEntityToWebsite.php
+++ b/src/Traits/ConvertsEntityToWebsite.php
@@ -20,19 +20,19 @@ use Hyn\Tenancy\Contracts\Website;
 trait ConvertsEntityToWebsite
 {
     /**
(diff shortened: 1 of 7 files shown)


src/Traits/ConvertsEntityToWebsite.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Hyn\Tenancy\Traits;
+
+use Hyn\Tenancy\Contracts\Hostname;
+use Hyn\Tenancy\Contracts\Website;
+
+trait ConvertsEntityToWebsite
+{
+    /**
+     * @param $to
+     * @return Hostname|Website|null
```
Thijmen: Should be `Website|null`, no `Hostname`.

src/Traits/ConvertsEntityToWebsite.php
```diff
@@ -0,0 +1,40 @@
...
+ * @see https://laravel-tenancy.com
+ * @see https://github.com/hyn/multi-tenant
+ */
+
+namespace Hyn\Tenancy\Traits;
+
+use Hyn\Tenancy\Contracts\Hostname;
+use Hyn\Tenancy\Contracts\Website;
+
+trait ConvertsEntityToWebsite
```
Thijmen, line 20: Maybe better name; ConvertHostnameOrWebsiteToWebsite? "Entity" seems so common. It doesnt convert everything.

tests/unit-tests/Queue/JobTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+        /** @var JobExtend $job */
+        $job = unserialize($attributes);
+
+        $this->assertInstanceOf(JobExtend::class, $job);
+
+        $restored = $this->environment->hostname();
+
+        $this->assertNotEquals($identified->fqdn, $restored->fqdn);
+    }
```
Thijmen: I'd like to see a test that injects some things in the database. Just to be sure. I like how you created this test, but I would be much comfortable if the data injected is also injected to the right database.

Thijmen (commented):
Here is my review, job well done, some small tweaks and we're there! :-)

src/Traits/ConvertsEntityToWebsite.php
luceos: It can also be a Hostname, hence the `instanceof` test a little down below.

src/Traits/ConvertsEntityToWebsite.php
luceos: I dislike class names of that length. I know it's not the most perfect trait name, but at least it covers the fact that it tries to retrieve a Website from an Entity.

tests/unit-tests/Queue/JobTest.php
luceos: I understand I would have loved to do the same. But.. to do a full fledged, real life test would require the Job process to be executed outside of the current process (to ensure no other tenant was previously activated). I could come up with a solution but the current test already proves the implementation works as is.

PR #334, opened 2017-11-21
Disable SQL-user generation

Thijmen:
Since I do not  want to create a SQL-user, I decided to create a PR that disables the SQL-user generation.

If a user of the package disables the generation of the SQL-user, he/she needs to make sure that the sql-user on the 'tenant'-connection can access the tenant database.

What are your thoughts?

### DIFF
--- a/assets/configs/tenancy.php
+++ b/assets/configs/tenancy.php
@@ -185,6 +185,13 @@ return [
         'password-generator' => Hyn\Tenancy\Generators\Database\DefaultPasswordGenerator::class,
 
         /**
+         * Standard, tenancy will create a SQL-user for you and grant it access to a new created database.
+         * With this setting, you can turn that off. Please note that the user and password of the tenant-connection
+         * will be used to connect to the database if turned off.
+         */
+        'generate-sql-user' => true,
+
+        /**
          * The tenant migrations to be run during creation of a tenant. Specify a directory
          * to run the migrations from. If specified these migrations will be executed
          * whenever a new tenant is created.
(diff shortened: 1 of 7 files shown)


tests/unit-tests/Repositories/WebsiteRepositoryTest.php
```diff
@@ -19,6 +19,23 @@
...
+
+    public function __construct()
+    {
+        parent::__construct();
+        $db = getenv('DB_CONNECTION');
+        if (!$db || $db == 'mysql') {
+            $this->tableForUsers = 'mysql.user';
+        } else {
+            $this->tableForUsers = 'pg_catalog.pg_user';
+        }
```
bkintanar: Can we remove else here by declaring `tableForUsers` before the `if` statement?

bkintanar (changes requested):

assets/configs/tenancy.php
```diff
@@ -184,6 +184,16 @@
          */
         'password-generator' => Hyn\Tenancy\Generators\Database\DefaultPasswordGenerator::class,
 
+        /**
+         * By default tenancy will automatically create a database user with access
+         * to the tenant database. By disabling this feature you are able to
+         * implement your own user generation or - even if not recommended -
+         * re-use existing credentials.
```
luceos, line 191: @Thijmen I've rewritten this comment. You mentioned it uses the tenant connection, but in fact it clones the system connection config, not the tenant one? If so it should read:

"re-use existing system connection credentials"

luceos:
@Thijmen we talked about creating DatabaseCredentialsGenerators where we will have one for separate databases, one database separate tables etc. These generators will have to share some information, eg do we need to create a database/username etc..

Closing this one.

PR #346, opened 2017-12-10
Invalid Hostname detection broken

fletch3555:
**Background:**
I was having significant difficulty getting my site to 404 on invalid hostnames, and found a few issues like missing config keys and other easily resolved issues.  Also found that the `HostnameActions::abort()` method was never firing.  After a little digging, I discovered the issue.

**Problem:**
If Eager (early) Identification is enabled, it will grab the `Environment`, which (may) auto-identify the current hostname.  The `$hostname` passed into the `HostnameActions` constructor is a blank object, since it's not tied to anything.  What we REALLY need here is `CurrentHostname`, but we can't DI that there because it returns an implementation of `Hostname`, not of `CurrentHostname`.

**Solution:**
I removed `$this->hostname` since it was passed directly into all other methods anyway, and made changed it to use `CurrentHostname` instead.  This

### DIFF
--- a/src/Middleware/HostnameActions.php
+++ b/src/Middleware/HostnameActions.php
@@ -15,6 +15,7 @@
 namespace Hyn\Tenancy\Middleware;
 
 use Closure;
+use Hyn\Tenancy\Contracts\CurrentHostname;
 use Hyn\Tenancy\Contracts\Hostname;
 use Hyn\Tenancy\Events\Hostnames\NoneFound;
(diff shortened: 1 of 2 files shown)


luceos:
I came across this issue today as well. I'll look at this in the morning. Thanks for sending it in!

luceos:
@fletch3555 thanks Eric, tagged 5.0.17

PR #352, opened 2017-12-20
Tenant language improvements and fixes

ianlchapman:
Fixed a bug where language files in a tenants directory were not being loaded if language files already exist in the global project.

Changed behaviour of tenant language files to merge with, rather than override entirely, global language files.


ianlchapman:
Code quality issue

PR #353, opened 2017-12-20
Improvements to tenant language functionality

ianlchapman:
Fixed bug where tenant language files were not being updated if global language files exist

Added functionality for tenant language files to extend and override rather than replace global language files.

### DIFF
--- a/src/Translations/MultiFileLoader.php
+++ b/src/Translations/MultiFileLoader.php
@@ -12,7 +12,7 @@
  * @see https://github.com/hyn/multi-tenant
  */
 
-namespace Hyn\Tenancy\Listeners\Filesystem;
+namespace Hyn\Tenancy\Tranlsations;
 
 use Illuminate\Translation\FileLoader;
 use Illuminate\Contracts\Translation\Loader;
--- a/src/Translations/MultiFileLoader.php
+++ b/src/Translations/MultiFileLoader.php
@@ -12,7 +12,7 @@
  * @see https://github.com/hyn/multi-tenant
  */
 
-namespace Hyn\Tenancy\Tranlsations;
+namespace Hyn\Tenancy\Translations;
 
 use Illuminate\Translation\FileLoader;
 use Illuminate\Contracts\Translation\Loader;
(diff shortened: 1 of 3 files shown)


src/Translations/MultiFileLoader.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Create a new file loader instance.
+     *
+     * @param  \Illuminate\Filesystem\Filesystem  $files
+     * @param  string  $path
+     * @return void
+     */
+    public function __construct(Filesystem $files, $path)
+    {
+        parent::__construct($files, $path);
```
luceos, line 37: There's no need to define the construct method if all it does is proxy to the parent.

src/Translations/MultiFileLoader.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param  string  $locale
+     * @param  string  $group
+     * @param  string  $namespace
+     * @return array
+     */
+    public function load($locale, $group, $namespace = null)
+    {
+        $results = [];
+        foreach ($this->fileLoaders as $loader) {
+            $results = array_merge($results, $loader->load($locale, $group, $namespace));
```
luceos, line 61: I think you should create a test that confirms the precedence of the existing global translation over the tenant translation. From what I read here, any subsequent loader will override the parent.

luceos (changes requested):
I've added a few comments and asked @Thijmen to do a review too.

Sorry to be a bother, but this is a pretty important part of the code base ;) Well done btw.

src/Translations/MultiFileLoader.php
Thijmen: I agree, I would like to see a test for this.

luceos:
@ianlchapman would you have time on your hands for that last hurdle? Please let me know if you need assistance.

Thijmen:
@ianlchapman Can you write some tests as reviewed above?

bkintanar:
@ianlchapman let us know if you can write some tests on this so we could review.

luceos:
@Thijmen @bkintanar let's do this ourselves. If either of you is up for it, assign the pr to yourself. Otherwise I'll pick it up within a week.

PR #360, opened 2018-01-11
Use `exists()` method instead of `exists` property on repositories

AustinMaddox:
The create() and update() methods on the repositories don't seem to be behaving as they should.

I'm not positive, but shouldn't these be `exists()` methods instead of properties? Changing them to methods seemed to fix the problem I have was having when creating things that already exist.

### DIFF
--- a/src/Repositories/CustomerRepository.php
+++ b/src/Repositories/CustomerRepository.php
@@ -52,7 +52,7 @@ class CustomerRepository implements Contract
      */
     public function create(Customer &$customer): Customer
     {
-        if ($customer->exists) {
+        if ($customer->exists()) {
             return $this->update($customer);
         }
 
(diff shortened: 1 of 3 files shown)


luceos:
@AustinMaddox thank you so much, clearly a stupid oversight 👍

Thijmen:
$user->exists contains true if the model is the real model taken from the database. For example, if you create $user = new User and set some properties to it, $user->exists will contain false, as long as you don't do $user->save(). Once it's saved or obtained from database, $model->exists will be true.

On the other hand, $user->exists() is a builder's method. It runs SQL's EXISTS query. If you run $user->exists(), you will generate a query:

TLDR; $object->exists is the way to go.

AustinMaddox:
Ah... after reading @Thijmen's explanation I agree with him. Disregard/close this PR.

luceos:
Reset the merge and tag.

PR #363, opened 2018-01-14
Update TenantAwareJob.php

Thijmen:
When I was using this trait, I found it annoying that the variable was named `tenant_id`, yet it is `hostname_id`.


fletch3555:
I'm also a fan of pedantry. 👍

luceos:
One drawback @Thijmen ; this is not backward compatible. We should tag 5.1 after.

luceos (approved):

Thijmen:
Yeah forgot about that. Let's tage 5.1? @luceos

Issue #357, opened 2017-12-27
I can't load any non tenant website e.g. admin dashboard


gkarugi:
Everything is working as expected with tenant websites. How can I set the system connection as the default on the fly when the base URL (base URL should hold the admin panel) is requested?

-Edited

Thijmen:
You could create a middleware for that, you can look into our middleware for identification hostnames and implement it as you want it :-)

gkarugi:
The tenant host names are the ones saved in the database, but the URL which should hold the main website and admin dashboard is not saved in the database. Can you point me in the right direction on how I can use a middleware to accomplish this?

luceos:
Can't your dashboard models not simply connect to the system database by applying that trait?

gkarugi:
it seems there is a global middleware which enforces the multitenancy. I have installed the application on azure. When I use their automatic deployment from a git repository tool, I am getting the following error :

> @php artisan package:discover

In Connection.php line 664:        
                                                         
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'masterdb.hostnam  
  es' doesn't exist (SQL: select * from `hostnames` where `fqdn` = uza.scm.az  
  urewebsites.net limit 1)      
                                               
In PDOConnection.php line 79:      
                                                                  
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'masterdb.hostnam  
  es' doesn't exist 

In PDOConnection.php line 77:

     Script @php artisan package:discover handling the post-autoload-dump event returned with error 
     code 1       
                                                                    
     SQLSTATE[42S02]: Base table or view not found: 1146 Table 'masterdb.hostnam  
    es' doesn't exist

gkarugi:
@luceos , I can successfully use the system connection trait.

I get the Exception Target [Hyn\Tenancy\Contracts\CurrentHostname] is not instantiable. when I try accessing any other FQDN that is not linked to a tenant.

Lets take a sub domains approach for creating tenants :

When a tenant is first created, their app is located on tenant1.domain.com, when someone tries accessing domain.com the Target [Hyn\Tenancy\Contracts\CurrentHostname] is not instantiable. exception is given.

luceos:
@gkarugi how is your tenant identification set up? Can you post your tenancy.php configuration?

gkarugi:
I have the set the following to env
` TENANCY_AUTO_HOSTNAME_IDENTIFICATION = false`

````
<?php
[...]

PR #365, opened 2018-01-20
Fix for CurrentHostname throwing error

fletch3555:
Related to/Partial fix for #357 and partially caused by [my fix](https://github.com/hyn/multi-tenant/pull/346/files#diff-0bfc9e44dd1452c380759a4f4e98ebe5R54) for #346 

Moved definition of CurrentHostname singleton to it's own ServiceProvider so requests for it don't trigger "cannot be instantiated" errors.  Defining singleton does not run it's closure, so no need to check for the `auto-identification` config.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -38,7 +38,6 @@ class Environment
         $this->app = $app;
 
         if ($this->installed() && config('tenancy.hostname.auto-identification')) {
-            $this->identifyHostname();
             // Identifies the current hostname, sets the binding using the native resolving strategy.
             $this->app->make(CurrentHostname::class);
         }
@@ -50,18 +49,6 @@ class Environment
     public function installed(): bool
     {
         return file_exists(base_path('tenancy.json'));
-    }
-
-    /**
-     * Auto identification of the tenant hostname to use.
-     */
-    public function identifyHostname()
-    {
-        $this->app->singleton(CurrentHostname::class, function () {
-            $hostname = $this->dispatch(new HostnameIdentification);
-
-            return $hostname;
-        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


luceos:
The drawback of this approach and also the reason why it requires the tenancy.json (used as a filesystem based flag) is to ensure migrations for tenancy have ran before attempting to run a Hostname Identification. Otherwise we'd have to catch errors thrown by the database whenever the hostnames table is not present.

fletch3555:
Just for reference, I believe #366/#367 would resolve your concern about this.  Do you agree?

luceos:
Had to merge locally, creating new pr for this. Thanks!

fletch3555:
@luceos, I don't see the changes in the codebase.  Did you actually merge this and push it? If so, where?

pvledoux:
@luceos any news for this PR? It's working locally for me.

luceos:
I've refactored several PR's into master taking the best of both worlds. I'm running into an issue right that needs fixing before I can tag a new minor update.

luceos:
@fletch3555 would you be able to review dev-5.x to see whether it solves this current issue?

fletch3555:
Just got another chance to look at things.  Looks like my HostnameProvider was dropped when you merged 5.x into this branch (commit ea649f9).  I feel like this was a mistake, though I don't see any newer minor/patch releases, so it's possible you're still working on it.  Please let me know

luceos:
@fletch3555 yes I've taken another approach at implementing what you suggested, I've been able to stabilize the 5.x-dev branch but we're holding off on tagging a release so we can consider adding Laravel 5.6 support into our 5.1 version. I'm still curious whether it works for you.

PR #367, opened 2018-01-24
Removed need for installation check by not query hostnames when running in cli

### LINKED ISSUE #366: What's the reason behind the tenancy.json file
Hi,

I'm just getting started with this package, but I'm finding it does pretty much everything I need :grinning:!

I am having some workflow issues though. I'm installing and running in a vagrant vm, however I tend to use artisan and composer on my host machine. Obviously, this doesn't work once installed (no database to connect to).

I'm wondering if the installation is really necessary? Assuming the following are true: 

- Installation is only used to determine if the hostname should be identified (no installation = no database)
- Installation is done through artisan
- Artisan should always default to the system connection

Rather than checking for an installation status to identify the hostname, surely it's just a case of setting the current hostname to null when running in the cli.

Happy to do a PR for this if agreed

Cheers,
George

PiranhaGeorge:
Based on discussions at #366 

Closes #366

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -35,7 +35,7 @@
         <env name="DB_DATABASE" value="tenancy"/>
         <env name="TENANCY_SYSTEM_CONNECTION_NAME" value="mysql"/>
         <env name="QUEUE_DRIVER" value="sync"/>
-        <env name="APP_ENV" value="production"/>
+        <env name="APP_ENV" value="testing"/>
         <env name="TENANCY_DEFAULT_HOSTNAME" value="local.testing"/>
     </php>
 </phpunit>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -35,7 +35,7 @@
         <env name="DB_DATABASE" value="tenancy"/>
         <env name="TENANCY_SYSTEM_CONNECTION_NAME" value="mysql"/>
         <env name="QUEUE_DRIVER" value="sync"/>
-        <env name="APP_ENV" value="testing"/>
+        <env name="APP_ENV" value="production"/>
         <env name="TENANCY_DEFAULT_HOSTNAME" value="local.testing"/>
     </php>
 </phpunit>
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -18,6 +18,7 @@ use Hyn\Tenancy\Contracts\CurrentHostname;
 use Hyn\Tenancy\Contracts\Customer;
 use Hyn\Tenancy\Contracts\Hostname;
 use Hyn\Tenancy\Contracts\Website;
+use Hyn\Tenancy\Database\Connection;
 use Hyn\Tenancy\Events\Hostnames\Switched;
 use Hyn\Tenancy\Jobs\HostnameIdentification;
 use Hyn\Tenancy\Traits\DispatchesEvents;
@@ -49,7 +50,9 @@ class Environment
      */
     public function installed(): bool
     {
-        return file_exists(base_path('tenancy.json'));
+        $connection = $this->app->make(Connection::class)->system();
+
+        return $connection->getSchemaBuilder()->hasTable('hostnames');
     }
 
     /**
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -60,14 +60,15 @@ class Environment
      */
     public function identifyHostname()
     {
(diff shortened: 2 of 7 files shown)


src/Environment.php
```diff
@@ -37,7 +38,7 @@ public function __construct(Application $app)
     {
         $this->app = $app;
 
-        if ($this->installed() && config('tenancy.hostname.auto-identification')) {
+        if (config('tenancy.hostname.auto-identification')) {
```
fletch3555, line 41: Where are we calling `installed()` after you remove it from here?  Does the method still need to exist, or was it a mistake to remove the call from this line?

src/Environment.php
```diff
@@ -58,9 +61,12 @@ public function installed(): bool
     public function identifyHostname()
     {
         $this->app->singleton(CurrentHostname::class, function () {
-            $hostname = $this->dispatch(new HostnameIdentification);
 
-            return $hostname;
+            if ($this->runningInConsole() || !$this->installed()) {
```
fletch3555: Only issue with doing it this way vs how I suggested in #366, is a very minor point of efficiency.  This method will make 1-2 queries (1 to check for table existence, and (possibly) 1 to identify the hostname).  The method I suggested will ALWAYS require a single query, and can just catch the "table not found" database exception and return null.

It's probably a moot point given how small the queries are.

src/Environment.php
PiranhaGeorge: Yeah, it's needed for tests, but not real http requests.

src/Environment.php
PiranhaGeorge: Good point.

PiranhaGeorge:
I don't like that the Environment is now 'test aware'.

I'm going to have a go at cleaning up some of the tests to fix this.

fletch3555:
I hate to be "that guy", but after one of your more recent commits, you're no longer calling `installed()`, again.

PiranhaGeorge:
@fletch3555 I am, I reverted that to calling it in controller.

I know your solution will work just fine, but it just doesn't feel right to query the DB when there's definitely not going to be a result.

phpunit.xml
```diff
@@ -35,7 +35,7 @@
         <env name="DB_DATABASE" value="tenancy"/>
         <env name="TENANCY_SYSTEM_CONNECTION_NAME" value="mysql"/>
         <env name="QUEUE_DRIVER" value="sync"/>
-        <env name="APP_ENV" value="production"/>
+        <env name="APP_ENV" value="testing"/>
```
Thijmen: This should be "production", because we need to test our code as it is running in production.

PiranhaGeorge: NP, This is not needed now that the tests have been updated.

tests/traits/InteractsWithTenancy.php
```diff
@@ -134,6 +135,18 @@ protected function setUpWebsites(bool $save = false, bool $connect = false)
         }
     }
 
+    protected function mockHttpRequest()
+    {
+        $this->app->singleton(Environment::class, function ($app) {
+            return new class($app) extends Environment {
+                public function runningInConsole()
```
fletch3555, line 142: Interesting solution.  Nice job

PiranhaGeorge: Thanks :-)

luceos: If I'm not mistaken the app is only instantiated once per test class, your ioc override will affect all tests inside the class once this method is executed. I'm pretty sure that's not intended.

PiranhaGeorge: I'll double check this, but I'm pretty sure it's not the case.

PiranhaGeorge: Okay, so createApplication() is invoked once in setUp() and then flushed during tearDown(). So, the application is created once, but the ioc container is flushed for each test. To confirm this I commented out the second call to mockHttpRequest() in InstallationTest and that test failed.

PiranhaGeorge: Actually, that's not quite right, createApplication() is called in every setUp().

luceos: Ah indeed, better safe than sorry 😊

Issue #383, opened 2018-02-12
Laravel 5.6 Dependancy


itguy614:
Update package for Laravel 5.6 support.

Thijmen:
It's on our roadmap, however feel free to create a PR which adds support for L5.6.

itguy614:
Forked and working through it.

itguy614:
Making some progress...

```
[02:16:57]$  vendor/bin/phpunit --configuration vendor/hyn/multi-tenant/phpunit.xml vendor/hyn/multi-tenant/tests/unit-tests/InstallationTest.php -vvv
PHPUnit 7.0.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.14
Configuration: /Users/kwolf/code/packages/multi-tenant/phpunit.xml

.........                                                           9 / 9 (100%)

Time: 484 ms, Memory: 16.00MB

OK (9 tests, 20 assertions)
```

itguy614:
Looking at the code from @Maveriks, I think he is now further along than I got to and I think his implementation is stronger than mine.  I'm going to close this in favor of the existing pull request.

pscheminger:
Wait, what do you mean it's on your "roadmap"? If you are going to put a package out for public use you should be one step ahead of the game and make sure it's usable on future Laravel releases. Am I wrong? I have a client using your package and it completely broke because you aren't supporting Laravel 5.6. Not good man, and it's put me in a very bad situation.

Thijmen:
@pscheminger You could've looked at our requirements (https://github.com/hyn/multi-tenant/blob/5.x/readme.md). It's not our fault it broke your app/website, as we clearly state it is only for 5.5 for now. We are busy on implementing 5.6, but as it is open source we are doing this in our free time.

However, feel free to put in a PR which supports 5.5 and 5.6, as we want to support the latest LTS of Laravel.

luceos:
@pscheminger 

Hello Phil,

As owner of this package I'd like to take my responsibility and point out my and your obligations.

- This package is open source. You are not paying for any level of support, there's no legal contract which requires me to provide it either.
- This package has a constraint on Laravel 5.5.* not 5.6.*. Composer would object installing it on laravel 5.6 or upgrading laravel to 5.6 while this package was installed.
[...]

PR #384, opened 2018-02-12
Feature/support 5.6

joshjacks:
Updated to the package to work with Laravel 5.6. This will break compatibility with 5.5.

### DIFF
--- a/assets/configs/webserver.php
+++ b/assets/configs/webserver.php
@@ -105,7 +105,7 @@ return [
         /**
          * The php sock to be used.
          */
-        'php-sock' => 'unix:/var/run/php/php7.0-fpm.sock',
+        'php-sock' => 'unix:/var/run/php/php7.1-fpm.sock',
 
         /**
          * Define the ports of your nginx service.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,14 +57,6 @@ references:
         - APP_KEY: deela5kinohw0haekoothahSh8eexach
 
 jobs:
-    "php-7.0-apache":
-        docker:
-            - image: php:7.0-apache
-              environment: *environment
-            - *mysql
-            - *mariadb
-            - *pgsql
-        <<: *steps
     "php-7.1-apache":
         docker:
             - image: php:7.1-apache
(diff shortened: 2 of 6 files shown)


luceos:
Tests are failing because L 5.6 needs 7.1+; this has to be modified inside the `.circleci/config.yml` file.

Question here is how and if we're going to support future versions. On discord Jiri mentioned supporting latest LTS and latest version of Laravel. Which makes a lot of sense.

assets/configs/webserver.php
```diff
@@ -105,7 +105,7 @@
         /**
          * The php sock to be used.
          */
-        'php-sock' => 'unix:/var/run/php/php7.0-fpm.sock',
```
luceos, line 108: good catch 👍

luceos:
Relates and closes #383 , #384

Thijmen:
Is it also possible to run the tests on L5.5? I have no experience with maintaining several laravel versions..

luceos:
There are several ways to maintain multiple versions.

- Do not support older versions, this is something @spatie is very good at 😁 I don't want to do this because creating an implementation (saas) can take quite a long time, it would feel horrible to have finished your app only to find this package is no longer patching issues.
- Maintain and work on both versions inside the same branch. This requires conditionals and other "hacks" which I seriously have a distaste for.
- Maintain and work on both versions using different versions, this is what I did before branches 3.x, 4.x and 5.x for laravel 5.3, 5.4 and 5.5. Getting new stuff in is a nightmare in this scenario.
- Work on the latest version and patch the older version. When a serious issue is found we can check out a tag, patch and tag. All new functionality and overhauls are done on the master (or primary branch). This will by default make all existing implementation out of date, I'm not too happy about this either.

Multi version support in all scenarios require a lot more work and thus time. With the current patron donations I still cannot dedicate enough time to work on tenancy at a steady pace. I wish this was different though.

 My current approach has been to support only the latest LTS version (hence dropping branches 3.x and 4.x from the scope completely, I even deleted all tagged versions).

dahabit:
waiting that pull request to use is with laravel 5.6 new project

luceos:
@Thijmen is going to take care of adding 5.6 support with backward compatibility. This will be part of the upcoming 5.1.0 release.

luceos:
@Maveriks, Josh thank you for doing the necessary work to get this in. Good job, much appreciated!

Issue #385, opened 2018-02-12
Tenancy database RefreshCommand migrates the system migrations in the websites


Dayvisson:
__Description__

The command **tenancy:migrate:refresh** runs:

- **migrate:reset** and   **migrate** 

 instead

-  **tenancy:migrate:reset** and **tenancy:migrate**

---

__Actual behavior__

This migrates the system migrations in the websites.

__Expected behavior__

The problem is the trait **MutatesMigrationCommands**  calls the parent:handle()  that is migrate:refresh that calls the defaults  **migrate:reset** and   **migrate**.

We should build a custom handle for Hyn\Tenancy\Database\Console\Migrations\RefreshCommand that calls the **tenancy:migrate:reset** and **tenancy:migrate** command.

I will try to send a PR to solve this.

---
__Information__

- hyn/multi-tenant version: 5.0.17
- laravel version: 5.5.25
- database driver and version: mysql 5.7
- webserver software and version: nginx 1.13.3
- php version: 7.1.11

luceos:
Thanks that must have been an oversight. Would love to see a PR!

Dayvisson:
Well, I'm working here and I may have found an alternative way to fix this issue.

Cause, if we choose rewrite the handle method from **Illuminate\Database\Console\Migrations\RefreshCommand** we have also to rewrite the methods **runRollback** and **runReset** and I don't know if this is the best way.

So, I rewrote just the handle method from the trait MutatesMigrationCommands, addind the path option. This way we're still runing the migrate:reset and migrate commands, but now  looking in the right paths.

I'll send the pull request now, the you can show this alternative.

luceos:
Will be part of 5.1.3.

PR #386, opened 2018-02-12
Fix tenancy:migrate:refresh bug - Issue#385

Dayvisson:
This PR solve the [issue#385](https://github.com/hyn/multi-tenant/issues/385).

I rewrote just the handle method from the trait **MutatesMigrationCommands**, adding the path option. This way we're still running the **migrate:reset** and **migrate** commands, but now looking in the right paths.

### DIFF
--- a/src/Database/Console/Migrations/RefreshCommand.php
+++ b/src/Database/Console/Migrations/RefreshCommand.php
@@ -20,4 +20,32 @@ use Illuminate\Database\Console\Migrations\RefreshCommand as BaseCommand;
 class RefreshCommand extends BaseCommand
 {
     use MutatesMigrationCommands;
+
+    public function handle()
+    {
+        if (!$this->confirmToProceed()) {
+            return;
+        }
+
+        $this->input->setOption('force', true);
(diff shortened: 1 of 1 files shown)


luceos:
Can you explain why the `getRelativeMigrationsPaths` is necessary?

Dayvisson:
The paths availables from MutatesMigrationCommands  (path, realpath or config('tenancy.db.tenant-migrations-path')) are all absolutes, but the command migrate:refresh should receive a relative path.

This way, I prefer do not loose the functionalities provides by getMigrationPaths() that you write, then I just transforms the paths returned by this methods in relatives ones and pass to the migrate:refresh command.

What do you think?

luceos:
The drawback of your method is that migrations outside the path of the app will no longer work. I'm not sure this is something we want. I also think this lack for absolute path is the reason I never implemented the Refresh Command in the first place. Laravel 5.6 might solve this with its native realpath support..

Dayvisson:
I get it. I'll implement another alternative rewriting the laravel's Refresh Command.

PR #387, opened 2018-02-13
Added onHostname method to TenantAwareJob.

joshjacks:
This method allows for the tenant the job is aware of to be manually overridden or set without an environment being set.

This is also useful for when jobs are going to be dispatched across many tenants. E.g. From a cron job can now easily dispatch a job on many tenants without having to set the environment.

### DIFF
--- a/src/Queue/TenantAwareJob.php
+++ b/src/Queue/TenantAwareJob.php
@@ -37,7 +37,7 @@ trait TenantAwareJob
 
         $hostname = $environment->hostname();
 
-        if (isset($hostname) && !isset($this->hostname_id)) {
+        if ($hostname && !$this->hostname_id) {
             $this->hostname_id = $hostname->id;
         }
 
@@ -62,7 +62,8 @@ trait TenantAwareJob
     /**
      * Manually override the hostname to be used.
      *
-     * @param mixed $hostname
+     * @param Hostname|int $hostname
+     * @return $this
      */
     public function onHostname($hostname)
     {
--- a/src/Queue/TenantAwareJob.php
+++ b/src/Queue/TenantAwareJob.php
@@ -35,7 +35,9 @@ trait TenantAwareJob
         /** @var Environment $environment */
         $environment = app(Environment::class);
 
-        if ($hostname = $environment->hostname()) {
+        $hostname = $environment->hostname();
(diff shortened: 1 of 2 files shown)


src/Queue/TenantAwareJob.php
```diff
@@ -35,7 +35,9 @@ public function __sleep()
         /** @var Environment $environment */
         $environment = app(Environment::class);
 
-        if ($hostname = $environment->hostname()) {
+        $hostname = $environment->hostname();
+
+        if (isset($hostname) && !isset($this->hostname_id)) {
```
luceos: I have a slight dislike for isset. Please use `if ($hostname && !$this->hostname_id)` instead.

Sorry matter of taste, we can safely assume the value to be null or set anyway.

src/Queue/TenantAwareJob.php
```diff
@@ -56,4 +58,16 @@ public function __wakeup()
 
         $this->serializedWakeup();
     }
+
+    /**
+     * Manually override the hostname to be used.
+     *
+     * @param mixed $hostname
```
luceos: I'm not a fan of mixed. Proper type hinting is crucial to get a good experience in your IDE. I agree we should allow both the Hostname object and an Id to be used as argument, if that's the case make sure the phpdoc matches.

Also this phpdoc is missing a `@return`.

tests/unit-tests/Queue/TenantAwareJobTest.php
```diff
@@ -61,4 +61,27 @@ public function serializes_tenant()
...
     }
+
+    /**
+     * @test
+     */
+    public function serializes_manual_tenant()
+    {
+        $this->app->make(CurrentHostname::class);
+
+        $manualHostname = $this->getReplicatedHostname();
```
luceos: There's no reason to call this variable manual at all. As you're not using the identified hostname, please use `$hostname` as variable.

luceos (changes requested):
Love the idea, great job. Added some thoughts I think make the code more in-line with the rest of the package.

🤗

luceos (approved):

luceos:
Thank you! Will be taken along with 5.1.

PR #388, opened 2018-02-15
Fix issue#385 rewriting the handle method

Dayvisson:
This is another alternative to solve the [issue#385](https://github.com/hyn/multi-tenant/issues/385) without any drawbacks.

I rewrite the handle method to execute the tenancy commands adding **website_id** and **realpath** params.


luceos:
Will pick this up asap, unless @Thijmen does.

bkintanar:
@luceos I think this needs attention. 😄

PR #389, opened 2018-02-16
Use trailing slash for Apache alias target

clarkwinkelmann:
Without a trailing slash all files in the aliased folder seem to return a 404 error. Adding the slash fixes the issue.

Not sure if the slash should be added in ApacheGenerator or in the template. Given paths are usually passed without a trailing slash in Laravel I think it makes sense to append it at the last moment.

This PR was tested on Ubuntu 16.04, Apache/2.4.18. I wonder why nobody else hit this issue ? Was nobody using the alias feature or does it work without the trailing slash on other systems ?

### DIFF
--- a/assets/generators/webserver/apache/blocks/server.blade.php
+++ b/assets/generators/webserver/apache/blocks/server.blade.php
@@ -8,7 +8,7 @@
 
     @if($media)
         # media directory
-        alias "/media/" "{{ $media . DIRECTORY_SEPARATOR }}"
+        alias "/media/" "{{ $media }}/"
     @endif
 
     # allow cross domain loading of resources
--- a/assets/generators/webserver/apache/blocks/server.blade.php
+++ b/assets/generators/webserver/apache/blocks/server.blade.php
@@ -8,7 +8,7 @@
 
     @if($media)
         # media directory
-        alias "/media/" "{{ $media }}"
+        alias "/media/" "{{ $media . DIRECTORY_SEPARATOR }}"
     @endif
 
     # allow cross domain loading of resources


assets/generators/webserver/apache/blocks/server.blade.php
```diff
@@ -8,7 +8,7 @@
 
     @if($media)
         # media directory
-        alias "/media/" "{{ $media }}"
+        alias "/media/" "{{ $media . DIRECTORY_SEPARATOR }}"
```
luceos: There's no real need to use DS here, the alias itself already holds a forward slash, let's stick to that. If you ammend that I'll gladly merge and bring in for 5.1.0

luceos (changes requested):
Hi Clark,

Thanks for pr'ing this. One minor edit required.

assets/generators/webserver/apache/blocks/server.blade.php
fletch3555: But that changes it back to the original state, and this is the only change in this PR... (o.0)

luceos: @fletch3555 `alias "/media/" "{{ $media "}}/"`

fletch3555: AH, see that makes much more sense. 👍

clarkwinkelmann:
I thought Laravel was mutating the slashes automatically for Windows with the path helpers, but it doesn't look like it is after all. So yes, unless you change the other method creating the media path it doesn't make sense to use DIRECTORY_SEPARATOR here.

luceos:
Thanks @clarkwinkelmann 👍

PR #410, opened 2018-04-06
support for multi dbs

luceos:
Currently we assume databases are created based off of the current system connection. Whenever you plan on moving a tenant to another server we need to allow the database creation there with full support to manage the database; including deletion and updating.

In order to do so we will be storing a column on the Website table which will relate to a database connection configured in `app/database.php`.

/cc @julesbloemen


Thijmen:
Does this mean it'll also use a new user, or does it uses the user defined in the connection?

luceos:
@Thijmen it uses any connection specified in the configuration. As this is now freely selectable during website creation it allows for use of any (secondary) system connection.

PR #433, opened 2018-04-25
Moving CurrentHostname resolution to it's own ServiceProvider.

fletch3555:
This is in relation to a conversation that occurred on Discord.

Currently, there are instances where calling `app(CurrentHostname::class)` will fail.

This pull request moves the hostname resolution logic to it's own ServiceProvider, which further decouples it from the rest of the code.  This means you are no longer required to resolve `Environment` first.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -41,8 +41,6 @@ class Environment
         $this->app = $app;
 
         if ($this->installed() && config('tenancy.hostname.auto-identification')) {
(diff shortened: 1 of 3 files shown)


fletch3555:
CircleCI is failing due to some `caching_sha2_password` plugin not being found in mysql.  This is unrelated to my changes.

luceos:
The `caching_sha2_password` error is caused by the release of MySQL 8. I'll look into it.

PR #475, opened 2018-05-30
[Fix] Seed command is not taking the option seeder (#474)

### LINKED ISSUE #474: Seed command is not taking the option seeder
The conditional is failing and therefor the default seeder is used

https://github.com/hyn/multi-tenant/blob/5.x/src/Traits/MutatesSeedCommands.php#L54

bkintanar:
Fixes #474

### DIFF
--- a/src/Traits/MutatesSeedCommands.php
+++ b/src/Traits/MutatesSeedCommands.php
@@ -51,7 +51,7 @@ trait MutatesSeedCommands
         $this->input->setOption('force', true);
         $this->input->setOption('database', $this->connection->tenantName());
 
-        if (! $this->option('class')) {
+        if (config('tenancy.db.tenant-seed-class')) {
             $this->input->setOption('class', config('tenancy.db.tenant-seed-class'));
         }
 


src/Traits/MutatesSeedCommands.php
```diff
@@ -51,7 +51,7 @@ public function handle()
         $this->input->setOption('force', true);
         $this->input->setOption('database', $this->connection->tenantName());
 
-        if (! $this->option('class')) {
+        if (config('tenancy.db.tenant-seed-class')) {
```
fletch3555, line 54: This doesn't allow users to change the seeder class via the command.  You still need to check for the `class` option somehow.

luceos: I agree with Eric, the root cause to fix is identifying why `$this->option('class')` is not a boolean false or null when you leave out `--class` while calling the command.

luceos:
The root cause of the seeder not taking the configured option is that the Laravel Seed Command has a default value for `--class` which is `DatabaseSeeder`. I'm pushing a fix for this.

I've been able to find the cause by adding tests and then running these tests singled out using the `--filter` option in phpunit. I dumped the option `class` and noticed the DatabaseSeeder string as value all the time.

PR #478, opened 2018-06-01
configure host when deleting.

bogdankharchenko:
noticed that this would error out with a undefined variable, 'host' when deleting website


luceos:
Thanks what a nice catch 👌

Could you add the same statement to the update method?

bogdankharchenko:
👍

luceos:
Much appreciated, will tag soon.

PR #516, opened 2018-06-22
Migrate Fresh Command

mintunitish:
This will add the command `tenancy:migrate:fresh` in the package.


mintunitish:
@luceos Please go through with this and suggest what I am missing.

PR #520, opened 2018-06-22
patched the fresh command

luceos:
@mintunitish so a few considerations:

- The `AddWebsiteFilterOnCommand` trait auto-added by the `MutatesMigrationCommands` added an option `website_id` expected to be an array; the `findOrFail()` call to the `websites` repository therefor would return a collection.
- The test ran `seedAndTest` but you should take care of seeding with the Fresh command; so I've reverted that to `migrateAndTest` and allowed that method to accept command options, so we can give it the `seed` and `seeder` options.
- The test had no existing websites to run any migrations on, so I've added a `setUpWebsites(true)` call.

Hope this clarifies the confusion.

### DIFF
--- a/src/Database/Console/Migrations/FreshCommand.php
+++ b/src/Database/Console/Migrations/FreshCommand.php
@@ -2,8 +2,7 @@
 
 namespace Hyn\Tenancy\Database\Console\Migrations;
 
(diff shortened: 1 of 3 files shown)


mintunitish:
Now I got it. Thanks for the review. I learned a lot!

PR #524, opened 2018-06-26
Add tenant-aware logger

janiskelemen:
I have created a tenant-aware logger with the following behavior:

If a valid hostname is detected it will create a _logs_ folder within the tenant directory ie.:
**app/tenancy/tenants/xxxx-xxxxxxxx-xxxx-xxxx/logs/{log_level}_xxxx-xx-xx.log**

If no valid hostname is detected it will default to:
 **/storage/logs/{log_level}_xxxx-xx-xx.log**

In order to activate the logger, you need to add the tenant channel in **config/logging.php** and change or add the stack.channels from single to tenant.
```
...
'channels' => [
        'stack' => [
            'driver' => 'stack',
            'channels' => ['single', 'tenant'],
        ],
        'tenant' => [
            'driver' => 'custom',
            'via' => Hyn\Tenancy\Logging\TenantAwareLogger::class,
            'level' => 'debug',
        ],
...
```

Hope this helps!

Best Regards
Janis

### DIFF
--- a/src/Logging/TenantAwareLogger.php
+++ b/src/Logging/TenantAwareLogger.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Hyn\Tenancy\Logging;
+
+use Monolog\Logger;
+use Illuminate\Support\Carbon;
+use Hyn\Tenancy\Website\Directory;
+use Monolog\Handler\StreamHandler;
+
+class TenantAwareLogger
+{
+    /**
+     * Create a custom Monolog instance and pipe logs to the tenant directory.
+     *
+     * @param  array  $config
+     * @return \Monolog\Logger
+     */
+    public function __invoke(array $config)
+    {
+        $log = new Logger('tenant');
+        $level = $log->toMonologLevel($config['level'] ?: 'debug');
+        $tenantDirectory = app(Directory::class);
+        $directoryPath = $tenantDirectory->getWebsite() ? 'app/tenancy/tenants/' . $tenantDirectory->path() : null;
+
+        $logPath = storage_path($directoryPath . 'logs/' . $config['level'] . '_' . Carbon::now()->toDateString() . '.log');
+        $log->pushHandler(new StreamHandler($logPath, $level));
+
+        return $log;
+    }
+}


janiskelemen:
Forgot to mention that the custom logger requires Laravel 5.6+

luceos:
I'm thrilled about the concept!

A few thoughts;

- the package also [requires Laravel 5.6](https://github.com/hyn/multi-tenant/blob/e7b0e3b15489d452e958dd9458dd6dc030b93253/composer.json#L30), all good there.
- requiring to configure the `logging.php` configuration file seems to be too much of a burden; the package configures database.php during runtime as well. As such it should be possible to;
  - create a new configuration key under `tenancy.folders` called `logging` with at least `enabled` defaulted to `false` and probably the same keys as the Laravel channel.
https://github.com/hyn/multi-tenant/blob/e7b0e3b15489d452e958dd9458dd6dc030b93253/assets/configs/tenancy.php#L299-L312
  - then we need another [Filesystem Listener](https://github.com/hyn/multi-tenant/tree/5.x/src/Listeners/Filesystem) `ConfiguresLogging` that reads the `enabled` flag and if set sets the Laravel `logging.channel` configuration and possibly update the `logging.default` as well as enforce/rebind that channel in case the Listeners was fired after the logger was set up.

This is a tad more work, but it provides a more transparent method compared to previous implementations.

janiskelemen:
Hi, thanks for your feedback. I'll try to add your suggestions within the next days.

luceos:
@janiskelemen thanks for your patience, and for taking the time to contribute; it's very much appreciated ❤️

luceos:
We found that making the default logger tenant aware caused a lot of issues. The LogManager in addition isn't too extensible. I am unsure how to improve the situation.

Some thoughts:

- Bind a new logger instance into the container.
- Create a custom TenantAwareLogger.
- Completely replace the Laravel Log manager.
- Register a new driver, which behinds the scenes, would be able to use any other driver..

janiskelemen:
@luceos I spent another few hours on this and came to the conclusion that changing the default LogManager is a probably too much hassle, just to get around the one-time config inside the logger file. I think we should leave the implementation to the original from this PR and let users decide if they want to add the tenant logger in their logging config. Another benefit is that we don't need to merge the custom logging drivers and channels since the order of the channels might be a concern at some point. Let me know what you think.

luceos:
@janiskelemen i was actually considering the same. As tenant directories already live in the local filesystem anyway.

~~Would you be able to PR against the docs or suggest what to write into it (summary) so we can merge and pick it up?~~ _see initial comment above_

I'm also asking @fletch3555 to do one last review of this additional code.

janiskelemen:
@luceos yes sure - I'll write the docs and make another PR on the docs repo.

fletch3555:
Code looks good to me.  Just a couple recommended changes to the docs, and this should be good to go.

PR #529, opened 2018-06-28
Use Website Contract instead of Models (#528)

nevertry:


luceos:
Oh jeez, what an oversight! Thanks so much!

PR #536, opened 2018-07-09
Dk/532 tenant routes

luceos:


src/Providers/Tenants/RouteProvider.php
```diff
@@ -17,31 +17,33 @@
...
+
+        if ($path) {
             /** @var Hostname $hostname */
             $hostname = $this->app->make(CurrentHostname::class);
 
             if ($hostname && file_exists($path)) {
-                if ($config->get('tenancy.routes.replace-global')) {
-                    $router->setRoutes(new RouteCollection());
-                }
+                $this->app->booted(function () use ($config, $router, $path) {
```
fletch3555, line 40: Hm, interesting fix... I was looking for something similar, but hadn't put much effort into it.  Nice job!  I'll give it a test later tonight or tomorrow.

luceos: @fletch3555 just out of transparency; I actually looked at both the Laravel Router and RouteServiceProvider (not the app one but the Illuminate one). It seems that the service provider adds a booted callback as well to refresh the routes (not to add, remove anything).

Issue #568, opened 2018-07-31
[Feature Request] Console command tenancy:run


athiffau:
__Description of your feature__

Version 2.x of Hyn/multi-tenant had a command called _multi-tenant:run_  which could be used to run any artisan command on a specific tenant.

One usage scenario is for confirming tenant specific routes with the artisan _route:list_ command.

---

__Proposed behavior__

For version 5.x of Hyn\Multi-tenant I suggest the command tenancy:run having the following signature:
```
protected $signature = 'tenancy:run {tenantcommand}
        {--tenant= : The tenant(s) to apply on; use all|5,8 }
        {--arguments= : Arguments for the delegated command}
        {--options= : Options to pass on to the delegated command}
    ';
```

__Current behavior__

The feature is currently not supported/implemented.

---
__Information__

- hyn/multi-tenant version: 5.x
- laravel version: 5.6

mintunitish:
@luceos Please look into the PR and suggest changes if required!

DeveloperOnCall:
@mintunitish I too need this (asap) so I merged your code into my versions - problem is, I get errors when using it. 

 **ReflectionException  : Method Hyn\Tenancy\Commands\RunCommand::handle() does not exist**

So I change the method fire() to handle() and am presented with this error:

**Call to undefined method Hyn\Tenancy\Repositories\WebsiteRepository::all()**

I then tried to adapt the code to work within the command I am trying to run and that didn't work either.

Any advice on getting this to work would be most appreciated.

Daniel

athiffau:
I tried the proposed PR and it does not "fully" work.  I had tried to port the 2.x command over as well and pretty much got the same result.

The issue I have is that some routes are missing when running `route:list` via this command.   I am comparing Laravel 5.6 with Hyn 5.1 to Laravel 5.3 with Hyn 2.x and not sure if this is due to Laravel changes.  The problem, I believe, is that with Laravel 5.6, the app isn't fully booted / bootstrapped when artisan runs.

I know that some routes are missing based on 1) the routes being accessible via URL and 2) installing this package: https://github.com/recca0120/laravel-terminal which show all of the routes described below.  

My setup customizes routes in three ways:

1 - conditional statements in routes\web.php
2 - tenant routes in storage\app\tenancy\tenants\XXXX\routes.php
3 - per tenant modules based on https://github.com/caffeinated/modules

Using tenancy:run with the code in this PR, 2 and 3 above are missing.  

[...]

PR #570, opened 2018-08-03
WIP : Tenancy Run Command

mintunitish:
#568


composer.json
```diff
@@ -31,9 +31,9 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~7.0",
-        "symfony/dom-crawler": "~3.1",
+        "symfony/dom-crawler": "~4.1",
```
fletch3555: These seem unnecessary.  Are the version updates required for this?

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Console\Kernel;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Hyn\Tenancy\Models\Website;
+
+class RunCommand extends Command
+{
+    protected $signature = 'tenancy:run {tenantcommand}
+        {--tenant= : The tenant(s) to apply on; use {all|5,8}}
```
fletch3555: Why not make this accept multiple values using built-in laravel functionality? https://laravel.com/docs/5.6/artisan#input-arrays

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+        {--options= : Options to pass on to the delegated command}
+    ';
+
+    protected $description = 'Run another artisan command in a tenant configuration';
+
+    public function fire()
+    {
+        $websites = $this->getWebsitesFromOption();
+        $newArgv = ['artisan', $this->argument('tenantcommand')];
+        if ($arguments = $this->option('arguments')) {
```
fletch3555: Why not use `$this->arguments()` then filter out the ones required by this command (i.e. `tenantcommand`)?

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+            $newArgv = array_merge($newArgv, explode(' ', trim($arguments)));
+        }
+        if ($options = $this->option('options')) {
```
fletch3555: Similar to above, why not use `$this->options()` then filter out the one(s) required by this command (i.e. `--tenant`)?

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+            $newArgv = array_merge($newArgv, explode(' ', trim($options)));
+        }
+        $this->output->progressStart(count($websites));
+        foreach ($websites as $website) {
+            putenv('TENANT=' . $website->id);
+            $tenantApp = require base_path('bootstrap') . '/app.php';
+            $kernel = $tenantApp->make(Kernel::class);
+            $status = $kernel->handle(
```
fletch3555: Since these are all artisan commands you'd run with this, why not use the Artisan facade to call it? https://laravel.com/docs/5.6/artisan#programmatically-executing-commands

src/Providers/TenancyProvider.php
```diff
@@ -14,6 +14,7 @@
 
 namespace Hyn\Tenancy\Providers;
 
+use Hyn\Tenancy\Commands\RunCommand;
```
fletch3555: I feel like these `use` statements should be alphabetical, but for the time being, could you at least group this command with the others below (lines 25-26)?

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * @return mixed
+     */
+    protected function getWebsitesFromOption()
+    {
+        $repository = app(WebsiteRepository::class);
+        if ($this->option('tenant') == 'all') {
+            return Website::all();
```
Namoshek: Using `$repository->query()->all()` would be better. Accessing `Hyn\Tenancy\Models\Website` directly will cause problems when a custom model is defined in `config('tenency.models.website')`.

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return $repository
```
Namoshek: There is no default action in place when no `--tenant=` option is given. `whereIn('id', [])` will lead to zero results. I think the command should raise an error in this case.

src/Commands/RunCommand.php
Namoshek: In the current version of Laravel, we can use `{--tenant=*}` to allow for multiple occurences of the option. This means we can call the command like `php artisan tenancy:run --tenant=1 --tenant=2 --tenant=7`. `$this->option('tenant')` will return `[1, 2, 7]` in this case.

Personally, I think the above is cleaner, but I see a point for your solution due to its brevity. It could be discussed whether `--tenants=` (plural) would be better, though, as the command description suggests two use cases where more than one tenant is addressed.

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+use Hyn\Tenancy\Contracts\Repositories\WebsiteRepository;
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Console\Kernel;
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Hyn\Tenancy\Models\Website;
+
+class RunCommand extends Command
+{
+    protected $signature = 'tenancy:run {tenantcommand}
```
Namoshek: Maybe change `{tenantcommand}` to `{command}` - I think it is clear what the argument stands for.

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,68 @@
...
+        $newArgv = ['artisan', $this->argument('tenantcommand')];
+        if ($arguments = $this->option('arguments')) {
+            $newArgv = array_merge($newArgv, explode(' ', trim($arguments)));
+        }
+        if ($options = $this->option('options')) {
+            $newArgv = array_merge($newArgv, explode(' ', trim($options)));
+        }
+        $this->output->progressStart(count($websites));
+        foreach ($websites as $website) {
+            putenv('TENANT=' . $website->id);
```
Namoshek: I don't think there is an environment variable called `TENANT`?!

Namoshek (changes requested):
Just some remarks...

mintunitish:
@Namoshek Some valid points there.

fletch3555 (changes requested):

src/Commands/RunCommand.php
mintunitish: Can you explain this a little bit please?

fletch3555: Instead of a single parameter on this command `--arguments`, you can just read the arguments as a collection with `$this->arguments()`.  This will return your `{command}` argument in the set, so you'll need to filter that value out.

After doing this, the command could be called like:
`artisan tenancy:run command arg1 arg2`
instead of
`artisan tenancy:run command --arguments=arg1,arg2`

src/Commands/RunCommand.php
mintunitish: Done!

src/Commands/RunCommand.php
mintunitish: And how the filtration will go?

src/Providers/TenancyProvider.php
mintunitish: Done!

src/Commands/RunCommand.php
fletch3555: That's really up to you I guess, but you could do something like:
```
$ignoredArgs = [$command, /* others */];
$args = array_filter($this->arguments(), function($arg) use ($ignoredArgs) {
    return !in_array($arg, $ignoredArgs);
});
```

src/Commands/RecreateCommand.php
```diff
@@ -89,7 +89,7 @@ protected function tenantDatabaseExists(Website $website) : bool
                 return true;
             }
         } catch (\Exception $e) {
-            // Surpress exception
+            // Suppress exception
```
luceos, line 92: For any following contributions, make sure not to mix these up with the actual PR. If a PR is not merged this kind of changes might be lost.

luceos (changes requested):
I'm not sure about your approach, here is why:

There is no reason to load the bootstrap just to get the Application instance, `app()` would suffice.


What I would do is using the `Environment` forcing the current environment by using the `tenant()` method and then running the command using the `$this->call()` functionality, see https://laravel.com/docs/5.6/artisan#calling-commands-from-other-commands

src/Commands/RunCommand.php
```diff
@@ -0,0 +1,74 @@
...
+    protected function getWebsitesFromOption()
+    {
+        $repository = app(WebsiteRepository::class);
+        if ($this->option('tenant') == 'all') {
+            return $repository->query()->all();
+        }
+
+        $tenantOptionArr = explode(',', $this->option('tenant'));
+
+        if (count($tenantOptionArr) == 0)
```
luceos: - Please use strict `===` checking with integers.
- Please use a shorter variable name, as `$tenants` is not used in this method, we can use that one.

luceos (changes requested):

src/Commands/RecreateCommand.php
mintunitish: Should I open a different PR for this?

luceos: No need 👍 this was just a pointer for next time.

luceos:
I tried some of your code and there were so many errors it was unusable. 

For instance the argument `command` can't be used. Symfony throws an error in that situation:

```
1) Hyn\Tenancy\Tests\Commands\RunCommandTest::can_proxy_artisan_commands
Symfony\Component\Console\Exception\LogicException: An argument with name "command" already exists.
```

I am going to give this a spin myself. My apologies, and thank you for giving this the attention it deserved! 🤗

Issue #575, opened 2018-08-09
Question: Maintance Mode not for admin area.


Nowi5:
Hi,

I would like to allow the tenant admin to de/activate the maintenance mode. The issue is that as soon he activates, he can't deactivate due to the fact that the admin interface isn't reachable, too. How can I extend / overwrite your Middleware/HostnameActions.php?

Thanks for your support.

luceos:
There are several solutions thinkable:

1. Extend the HostnameActions shipped with the package and enable it in your app (disable it in the tenancy.php). Then override the `maintenance` method to do whatever you want.
2. Use another property on the Hostname/Website to identify maintenance mode and built your own implementation.

Nowi5:
Thank you. You are right, the second seems to be the best idea. Due to performance, can I deactivate your middleware easily so that they do not need to check the DB?

luceos:
You can't disable/replace the complete Middleware; I'd gladly accept a PR for it though!

Nowi5:
What should be the goal of the PR?
1. Set a contract for the middleware class and allow to replace the middleware class by config?
2. Allow a whitelist of maintainance mode via config?
3. Allow via config the deactivation of "under_maintenance_since" function?

Without a PR I would just ignore this field, extend the model by my own field and attach a own middleware in addition to the existing one.

luceos:
Check this PR: https://github.com/hyn/multi-tenant/pull/579

Nowi5:
This was quick. So means more or less Option 1, we can replace the middleware. Thank you :-)

luceos:
The PR still needs to be modified and merged. After which a new version has to be released. I'll do so as soon as possible.

Nowi5:
Will do the update soon and let you know. Thanks.

luceos:
Middlewares are now configurable through the tenancy.php configuration file.

PR #579, opened 2018-08-14
Make middlewares customizable

yuters:
This allows to easily extend the package middlewares to implement custom logic and exceptions.

### DIFF
--- a/assets/configs/tenancy.php
+++ b/assets/configs/tenancy.php
@@ -30,16 +30,16 @@ return [
         // Must implement \Hyn\Tenancy\Contracts\Website
         'website' => \Hyn\Tenancy\Models\Website::class
     ],
+    /**
+     * The package middlewares. Removing a middleware here will disable it.
+     * You can of course extend/replace them or add your own.
+     */
     'middlewares' => [
-        /**
-         * The eager identification middleware.
-         */
-        'eager-identification' => \Hyn\Tenancy\Middleware\EagerIdentification::class,
+        // The eager identification middleware.
+        \Hyn\Tenancy\Middleware\EagerIdentification::class,
 
-        /**
-         * The hostname actions middleware (redirects, https, maintenance).
-         */
-        'hostname-actions' => \Hyn\Tenancy\Middleware\HostnameActions::class,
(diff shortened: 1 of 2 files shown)


luceos (changes requested):
Relates to #575 

@yuters thanks for this, that's really nifty. One request though; can you allow (and document in the config) that setting the config key to `null` will disable them? That would allow for the most flexibility.

yuters:
I was also thinking about the config not being an associative array, just a simple array of middlewares to prepend to the stack, so you can just remove the ones you don't need, or add to it. Do you think it's preferable?

luceos:
@yuters that crossed my mind too, but I didn't want to put too much pressure on your first contribution 👍 

So please take that path, it's possibly less overhead in terms of necessary code as well.

assets/configs/tenancy.php
```diff
@@ -30,6 +30,17 @@
         // Must implement \Hyn\Tenancy\Contracts\Website
         'website' => \Hyn\Tenancy\Models\Website::class
     ],
+    /**
+     * The package middlewares. Removing a middleware here will disable it.
+     * You can of course extend/replace them or add your own.
+     */
+    'middlewares' => [
```
fletch3555: One comment.  `middlewares` sounds weird to me (native-born English speaker).  Much like `deer`, `fish`, and even `software`, the pluralization of `middleware` is often just `middleware`.  Also, for reference: https://english.stackexchange.com/questions/257120/middleware-vs-middlewares

yuters: That’s a good point. I’m not a native English speaker.

What would you suggest to do in a foreach statement (ex. right now it’s foreach (middlewares as middleware)) to clearly distinguish the plural from the singular?

fletch3555: I was thinking about that earlier. I personally tend to use shorthand inside loops, so in my own code I'd probably use something like `mw` or `item`. But that's not necessarily the best option in a package like this.  I'll leave it up to you to come up with something reasonable.  Don't worry too much about singular vs plural (though that is a very good practice in general)

luceos:
Very happy with this, will see to a new tagged version asap!

PR #581, opened 2018-08-17
Allow custom/own mimes for Media controller

Nowi5:
Allow MIMES configuration for media folder via controller
as in some cases, especially shared host does not support the full list of mimes, here you can define all you like

### DIFF
--- a/assets/configs/tenancy.php
+++ b/assets/configs/tenancy.php
@@ -376,5 +376,16 @@ return [
              */
             'override-global' => true,
         ]
+    ],
+    /**
+     * MIMES configuration for media folder via controller
+     * In some cases especially shared host does not support the full list of mimes, here you can define all you like
+     * If you do not need this feature please comment
+     */
+    /*
+    'mimes' => [
+        'css'   =>  'text/css',
+        'js'    =>  'text/javascript'
     ]
+    */
 ];
--- a/src/Controllers/MediaController.php
+++ b/src/Controllers/MediaController.php
@@ -42,20 +42,14 @@ class MediaController
         $path = "media/$path";
 
         if ($this->directory->exists($path)) {
-
             $mimetype = Storage::disk('tenant')->mimeType($path);
 
             if (config('tenancy.mimes')){
                 // Unfortunately, some shared hoster have limited mime type support, so we need to improve that by our self.
                 $fileextension = pathinfo($path)['extension'];
 
-                $mimes = [
-                    'css' => 'text/css',
-                    'js' => 'text/javascript'
-                ];
-
-                if (isset($mimes[$fileextension])) {
-                    $mimetype = $mimes[$fileextension];
+                if (isset(config('tenancy.mimes')[$fileextension])) {
+                    $mimetype = config('tenancy.mimes')[$fileextension];
                 }
             }
 
--- a/src/Controllers/MediaController.php
+++ b/src/Controllers/MediaController.php
@@ -42,8 +42,25 @@ class MediaController
         $path = "media/$path";
 
         if ($this->directory->exists($path)) {
+
+            $mimetype = Storage::disk('tenant')->mimeType($path);
+
+            if (config('tenancy.mimes')){
+                // Unfortunately, some shared hoster have limited mime type support, so we need to improve that by our self.
+                $fileextension = pathinfo($path)['extension'];
+
+                $mimes = [
+                    'css' => 'text/css',
+                    'js' => 'text/javascript'
+                ];
+
+                if (isset($mimes[$fileextension])) {
+                    $mimetype = $mimes[$fileextension];
+                }
+            }
+
             return response($this->directory->get($path))
-                ->header('Content-Type', Storage::disk('tenant')->mimeType($path));
+                ->header('Content-Type', $mimetype);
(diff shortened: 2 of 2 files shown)


luceos:
This is a tough nut. Initially I was like "cool", but when I think about the underlying issue I can only suggest throwing errors when uploading instead?

Nowi5:
What do you mean with throwing errors when uploading? In my case, I generate for each tenant a own CSS file, after running with scss. This allows for each tenant a own theme. How should I solve that issue else? Okay I could forget about the tenant folder and generate my own tenant folder direct in the public section, with that I would not need the controller anymore. Any other idea?

athiffau:
If I understand the usage scenario this may help. In my case I wanted to support common files as well as per-tenant file access.  To accomplish this I implemented a custom dispatcher.  I look at the request and if the path matches my constraints I serve a file else I pass it on for Laravel to process. 

Examples:  

Base URL Request:
`https:\\<base.url>\assets\<path>` will serve a file from `\Resources\Assets\<path>`

Tenant URL Request:
`https::\\<tenant.base.url>\assets\<path>` will serve a file from `\Storage\app\tenants\<uuid>\assets\<path>`

Still, I wanted to reduce file duplication and have some common files so yet another constraint is:
`https:\\<tenant.base.url\cdn\assets\<path>` will serve a file from `\Resources\Assets\<path>`

I have several additional constraints that I check for supporting pulling files from modules and themes folders before I pass on the request to Laravel to serve from the `public` folder which is where the mix/webpack files are.

Nowi5:
Hi Athiffau,

the forwarding and the delivery of the file itself is not the challenge and working fine. The provided media provider is more or less doing what you said. The challenge is that if the PHP Server does not know the MIME Type of CSS and JS he will deliver text/plain instead of text/css. So that would happen in your script the same way. Without the right mime type browser will not except your CSS Code.

Thanks.

athiffau:
Right; I had that problem when I first implemented the dispatcher.  My implementation now works with limited testing that I have done.  I can serve CSS/JS files and the browsers accept it.

Here's a copy of how I do it.  Curious to hear what you think about this and especially if you think it doesn't cover all cases or problematic/unsafe. 

```
    /**
     * Serve a File.
     *
     * @param string $path
     *
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function serve($path, SymfonyRequest $request)
    {
        if (! file_exists($path)) {
            event(new InvalidFileRequest($request));
            return new Response('File Not Found', 404);
        } else if (! is_readable($path)) {
            return new Response('Unauthorized Access', 403);
        }

        // Retrieve the file content type.
        $mimeType = $this->getMimeType($path);

        // Check if the request support compression algorithms
        $algorithms = $this->getEncodingAlgorithms($request);

        // Determine if the file could be compressed.
        $compressable = (($mimeType == 'application/javascript') || str_is('text/*', $mimeType));

        if ($this->compressFiles() && ! empty($algorithms) && $compressable) {
            // Get the (first) encoding algorithm.
            $algorithm = array_shift($algorithms);

            // Retrieve the file content.
            $content = file_get_contents($path);

            // Encode the content using the specified algorithm.
            $content = $this->encodeContent($content, $algorithm);

            // Retrieve the Last-Modified information.
            $timestamp = filemtime($path);  //TODO: check if this means we bust cache everytime ?

            $modifyTime = Carbon::createFromTimestampUTC($timestamp);

            $lastModified = $modifyTime->format('D, j M Y H:i:s') .' GMT';

            // Create the custom Response instance.
            $response = new Response($content, 200, array(
                'Content-Type'     => $mimeType,
                'Content-Encoding' => $algorithm,
                'Last-Modified'    => $lastModified,
            ));
        } else {
            // Create a Binary File Response instance.
            $response = new BinaryFileResponse($path, 200, array(), true, 'inline', true, false);

            // Set the Content type.
            $response->headers->set('Content-Type', $mimeType);
        }

        // Setup the (browser) Cache Control.
        $this->setupCacheControl($response);

        // Prepare against the Request instance.
        $response->isNotModified($request);

        return $response;
    }
```

The getMimeType function is:
```
    protected function getMimeType($path)
    {
        // Even the Symfony's HTTP Foundation have troubles with the CSS and JS files?
        // Hard coding the correct mime types for presently needed file extensions.

        switch ($fileExt = pathinfo($path, PATHINFO_EXTENSION)) {
            case 'css':
                return 'text/css';

            case 'js':
                return 'application/javascript';

            case 'svg':
                return 'image/svg+xml';

            default:
                break;
        }

        // Guess the path's Mime Type.
        $guesser = MimeTypeGuesser::getInstance();

        return $guesser->guess($path);
    }
```

Nowi5:
Hi Athiffau, See you solve exactly the same issue I had with file extension and setting the mimetype manually. So we can use my PR so that not everybody has to do it by himself, right?

athiffau:
Sorry; didn't see that this was a PR - should of paid more attention. (👎 for me!) 
I definitely agree that this is valid and we should use common code.

luceos:
The media controller was shipped as an example. I don't think this package is responsible for doing smarter logic like filtering, white- or blacklisting etc. You're free to extend the controller or create your own.

PR #591, opened 2018-08-28
Add connection to table creation

Jeroen-G:
By specifying the `system` connection the tables are more obviously created on the system database.
I ran into this when using this package in an existing application, where it would have ran on a database that is not the system one.

### DIFF
--- a/assets/migrations/2017_01_01_000003_tenancy_websites.php
+++ b/assets/migrations/2017_01_01_000003_tenancy_websites.php
@@ -12,9 +12,9 @@
  * @see https://github.com/hyn/multi-tenant
  */
 
-use Hyn\Tenancy\Abstracts\AbstractMigration;
-use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Hyn\Tenancy\Abstracts\AbstractMigration;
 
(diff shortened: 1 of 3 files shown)


bkintanar:
I don't think this is necessary as the migration files for tenants will be run from the `database\migrations\tenant` directory. so all other migrations outside of this directory will be running with the system connection.

fletch3555:
@bkintanar not necessarily.  They will run on whatever the "default" connection is, which for some is `tenant`.

@Jeroen-G, you shouldn't use a hardcoded `"tenant"`.  Instead, you can get the system connection name from `app(Hyn\Tenancy\Database\Connection::class)->systemName()`

luceos:
> I ran into this when using this package in an existing application, where it would have ran on a database that is not the system one.

Out of curiosity @Jeroen-G did you actually ran into that issue?

Because the AbstractMigration in this package has a way of identifying the connection used, see: https://github.com/hyn/multi-tenant/blob/5.x/src/Abstracts/AbstractMigration.php

Jeroen-G:
I did, maybe because we not only had a system connection, but also a default tenant (while we were moving to a multitenant app).

luceos:
Did you set `database.default` to tenant? And that resulted in the migration being forced onto the tenant instead of the system?

luceos:
No valid reason yet, so closing.

PR #595, opened 2018-08-29
Disable bubble option in StreamHandler of TenantAwareLogger

janiskelemen:
Disable bubble option in StreamHandler of TenantAwareLogger in order to prevent empty brackets at the end of log lines.


luceos:
I should have merged this for 5.2.7, ah well :)

PR #599, opened 2018-08-31
add support for multiple tenant migration path

### LINKED ISSUE #598: multiple tenant migration path
it will be very great if we can do something like this

``` 
"tenant-migrations-path" => [
database_path('migrations/tenant1'), database_path('migrations/tenant2')
]
```

instead of just single path
```
"tenant-migrations-path" => database_path('migrations/tenant'),
```

vortechron:
Closes #598

### DIFF
--- a/src/Traits/MutatesMigrationCommands.php
+++ b/src/Traits/MutatesMigrationCommands.php
@@ -72,8 +72,8 @@ trait MutatesMigrationCommands
         }
 
         // Tenant migrations path is configured.
-        if ($path = config('tenancy.db.tenant-migrations-path')) {
-            return is_array($path) ? $path : [$path];
+        if ($path = (array) config('tenancy.db.tenant-migrations-path')) {
+            return $path;
         }
 
         throw new InvalidArgumentException("To prevent unwanted migrations from database/migrations, always specify a path.");
--- a/src/Traits/MutatesMigrationCommands.php
+++ b/src/Traits/MutatesMigrationCommands.php
@@ -73,7 +73,7 @@ trait MutatesMigrationCommands
 
         // Tenant migrations path is configured.
         if ($path = config('tenancy.db.tenant-migrations-path')) {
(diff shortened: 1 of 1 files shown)


luceos:
Wouldn't casting the value to array be easier?

vortechron:
@luceos more readable in my opinion :)

fletch3555:
Description edited to auto-close referenced issue when merged

luceos:
Laravel casts arguments to array everywhere. So if I have to choose between your personal preference and what the author of the framework we are working with does, then I'll go for the latter 😉

vortechron:
@luceos edited 👍

src/Traits/MutatesMigrationCommands.php
```diff
@@ -72,8 +72,8 @@ protected function getMigrationPaths()
         }
 
         // Tenant migrations path is configured.
-        if ($path = config('tenancy.db.tenant-migrations-path')) {
-            return [$path];
+        if ($path = (array) config('tenancy.db.tenant-migrations-path')) {
```
luceos: A little test in tinker:


```
>>> if ($var = (array) 'bar') { echo $var; }
ArrayPHP Notice:  Array to string conversion in Psy Shell code on line 2
>>> if ($var = (array) null) { echo $var; }
>>> if ($var = (array) '') { echo $var; }
ArrayPHP Notice:  Array to string conversion in Psy Shell code on line 2
>>>
```

Aka this implies that an empty string in the config already triggers the conditional. We would need to do an `empty()` check in the conditional and then cast the value to array on the return line.

luceos (changes requested):

luceos:
I will bug you once more 😞 Please reset to your initial proposal, it works out the best in this scenario.

Sorry for the troubles 🤗

vortechron:
@luceos haha its okay 👍

PR #602, opened 2018-08-31
Fr/600 use differents table names

ghost:
- Removed the term websites fixed in the code and obtained through the tenancy configuration file.
- Update to get table name by model instance.


luceos:
Thanks, I've closed #600 prematurely, but this is a great addition 👍

Issue #533, opened 2018-07-04
New Tenant Division Mode: Multischema for PostgreSQL


xmarkclx:
__Description of your feature__
Allow mode to use multischema on 1 database for PostgreSQL.

https://www.citusdata.com/blog/2016/10/03/designing-your-saas-database-for-high-scalability/ says why it is preferable.
..

---

__Proposed behavior__
Add a new tenant-division-mode option, along with codes to create the schema, query based on it, etc.

..

__Current behavior__
There is no such option.

..

---
__Information__

- hyn/multi-tenant version: 5.2.*
- laravel version: 5.6.*

luceos:
This would be great to have. But it doesn't have a high priority on my planning. A tested PR would be accepted.

fletch3555:
Closing as #603 got merged.

PR #603, opened 2018-09-03
Schema division mode

balping:
Implements feature described in #533 

I created a new DatabaseGenerator driver (`PostgresSchema`) that extends the default Postgres driver. To avoid code duplication, I refactored the postgres driver by extracting each step in creating/deleting databases into separate functions that can be overridden.

To test this new functionality, I added the following line to ci config:

```
DB_CONNECTION=pgsql DB_HOST=pgsql TENANCY_SYSTEM_CONNECTION_NAME=pgsql TENANCY_DATABASE_DIVISION_MODE=schema ./vendor/bin/phpunit -c ci.phpunit.xml ./tests/unit-tests/Database/ConnectionTest.php
```

There might be a better way of testing it, but this seems sufficient.

### DIFF
--- a/src/Generators/Webserver/Database/DatabaseGenerator.php
+++ b/src/Generators/Webserver/Database/DatabaseGenerator.php
@@ -66,7 +66,9 @@ class DatabaseGenerator
 
         switch ($driver) {
             case 'pgsql':
-                return $this->mode === Connection::DIVISION_MODE_SEPARATE_SCHEMA ? new Drivers\PostgresSchema : new Drivers\PostgreSQL;
+                return $this->mode === Connection::DIVISION_MODE_SEPARATE_SCHEMA 
+                    ? new Drivers\PostgresSchema 
+                    : new Drivers\PostgreSQL;
                 break;
             case 'mysql':
                 return new Drivers\MariaDB;
--- a/src/Generators/Webserver/Database/DatabaseGenerator.php
+++ b/src/Generators/Webserver/Database/DatabaseGenerator.php
@@ -66,6 +66,9 @@ class DatabaseGenerator
 
         switch ($driver) {
             case 'pgsql':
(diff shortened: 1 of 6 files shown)


src/Generators/Webserver/Database/DatabaseGenerator.php
```diff
@@ -66,6 +66,9 @@ protected function driver(array $config)
 
         switch ($driver) {
             case 'pgsql':
+                if ($this->mode === Connection::DIVISION_MODE_SEPARATE_SCHEMA) {
+                    return new Drivers\PostgresSchema;
+                }
```
luceos: Can we make the above a single line on the next return statement:

```php
return $this->mode === Connection::DIVISION_MODE_SEPARATE_SCHEMA ? new Drivers\PostgresSchema : new Drivers\PostgreSQL;

luceos (dismissed):
Very well done! I'm stunned. Just a minor nitpick from my side and it's good to go if you can guarantee you ran it locally.

src/Generators/Webserver/Database/DatabaseGenerator.php
balping: Originally I was going to do the same, but then I realised that it would be too wide for my taste, well beyond 80 columns, but even beyond 120 columns. But anyway, if you prefer the ternary style, it's fine too.

balping: And, yes I've run the whole test suite (with skips) and tested manually in a real laravel installation.

balping:
@luceos ping

fletch3555:
@balping, could you please fix the conflicts?

balping:
@fletch3555 done

luceos:
Thank you! 🤗

PR #605, opened 2018-09-04
Support newly released Laravel 5.7

richardkeep:


luceos:
If I'm not mistaken we can safely modify the constraint to be `^5.6.0` this would allow any future 5.x version?

fletch3555:
Yes, `^5.6.0` should be fine (and would allow `5.6.1`, `5.6.84`, `5.7.0`, `5.8.0`, etc).  The issue with this PR is that changing `5.6.0` to `5.7.0` removes support for 5.6, which is still widely used.

dbrw:
no release with this patch, yet?

richardkeep:
yep. :(

luceos:
@dbrw I'm working on it, there's just a minor change needed and I want to patch the tests before tagging a new version 👍

PR #612, opened 2018-09-08
Set Mysql Default Char Set and Collation on create

londoh:
I am working on a system where it will be helpful to be able to set Default Character Set and Default Collation by using data from database driver config. This pull request achieves that. I trust you find it acceptable. Otherwise maybe here is some method/setting I missed?

### DIFF
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -41,7 +41,9 @@ class MariaDB implements DatabaseGenerator
             return true;
         };
         $create = function ($connection) use ($config) {
-            return $connection->statement("CREATE DATABASE IF NOT EXISTS `{$config['database']}`");
+            return $connection->statement("CREATE DATABASE IF NOT EXISTS `{$config['database']}`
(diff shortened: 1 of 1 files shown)


fletch3555:
Doesn't Laravel automatically use the Charset and Collation from the DB connection config?

Assuming that's correct, you just need to set it on the `system` connection config and this package will automatically copy it to the `tenant` connection config.  No need to set it in the DB directly.

londoh:
Laravel most probably does  use charset and collation from db config. They are there as standard and are set correctly. in this installation. 
However when using multi-tenant, it is creating the new tenant DB, not laravel per se. 

In this instance the Mysql Server is outside of my control and the charset and collation are set to latin1, not as required for laravel defaults.

So I noticed each new database from multi-tenant was created with Server defaults - and not values from tenantl config. This patch appears to correct that.

I admit I didnt test extensively, and it may not make any difference as tables are created correctly. 
But given the option is there I'd prefer to have it correct at db level.

Apologies for not reading pr instructions.

luceos:
Thank you, I agree with the intent and the execution 👍

PR #617, opened 2018-09-15
add support for multiple tenant migration path

### LINKED ISSUE #598: multiple tenant migration path

vortechron:
closes #598

### DIFF
--- a/src/Listeners/Database/MigratesTenants.php
+++ b/src/Listeners/Database/MigratesTenants.php
@@ -47,14 +47,23 @@ class MigratesTenants
      */
     public function migrate(WebsiteEvent $event): bool
     {
-        $path = config('tenancy.db.tenant-migrations-path');
+        $paths = $this->getMigrationPaths();
 
-        if ($path && realpath($path) && $this->connection->migrate($event->website, $path)) {
-            $this->emitEvent(new Events\Websites\Migrated($event->website));
-
-            return true;
+        foreach ($paths as $path) {
+            if ($path && realpath($path) && $this->connection->migrate($event->website, $path)) {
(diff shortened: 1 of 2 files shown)


src/Listeners/Database/MigratesTenants.php
```diff
@@ -47,14 +47,23 @@ public function subscribe(Dispatcher $events)
...
+            }
         }
 
         return true;
     }
+
+    protected function getMigrationPaths()
+    {
+        if (($path = config('tenancy.db.tenant-migrations-path')) && ! empty($path)) {
+            return (array) $path;
```
fletch3555, line 64: Wouldn't this make it an array of arrays of paths? That doesn't seem to be what you're looking for.

vortechron: im not understand what u mean by that..

luceos: @fletch3555 this would only apply in case the configuration would be an array of empty strings or something; I think that would be very unlikely..

luceos:
Thanks, taking this along for the next release 👍

src/Listeners/Database/MigratesTenants.php
fletch3555: Apologies, you're correct.  I was thinking it was `return [$path]`, which would wrap whatever `$path` is in an array, which would make an array of arrays if that config value is already an array.

Instead, this is casting it as an array, which should be fine.

PR #636, opened 2018-10-12
Changed $exitCodes to a reference

Plytas:
Without this it always says `Command was executed on zero tenants.`

### DIFF
--- a/src/Commands/RunCommand.php
+++ b/src/Commands/RunCommand.php
@@ -73,7 +73,7 @@ class RunCommand extends Command
 
         $exitCodes = [];
 
-        $query->chunk(50, function ($websites) use ($environment, $options, $exitCodes) {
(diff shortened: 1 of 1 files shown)


mintunitish (approved):

Plytas:
What if we actually showed on how many tenants the command was run? Maybe even a progress bar while running?

fletch3555:
I like the progress bar idea.  Shouldn't be that difficult.

Plytas:
We could do `$query->count()` before chunking for the progress bar, but that would result in additional query. Is there a way to do it with just a single query?

luceos:
@Plytas good point, can you create an issue or PR for the count?

Please note I removed this from the original PR of the run command to lower the overhead. If there's a decent way to get it done nevertheless I'd love to hear so.

Issue #574, opened 2018-08-08
Allow APP_KEY rotation


luceos:
By @bkintanar ;)

Relates to https://t.co/XdhgIkbu0v

mintunitish:
There should be some mechanism that changes the APP_KEY automatically in some intervals.

luceos:
I think that having key rotation is very expensive with many tenants. In addition there are so many risks involved; eg jobs/crons/pages not loading due to the credentials being invalid. Something like this is best run during a maintenance window on larger platforms.

My suggestion is to write a `tenancy:key:rotate` command that will allow the key to be regenerated, and stored only after all tenant database user passwords have been updated. Something like this could be done faster by forking several processes, which in turn might run into concurrent connections issues with the system connection...

bkintanar:
I'll try to have a whack at it.

fletch3555:
This seems like such a one-off case... The issue is "resolved" by laravel disabling serialization.  The only reason to change keys is if you know it's been compromised.

Can we simply provide a method that accepts both old and new keys and handles it.  Then let the user create the command if that's how they want to do it, perhaps with an example in the docs.

mintunitish:
Agree with the idea of providing a command or some sort of method to do that. However it's being fixed in Laravel v5.6.30 so that's done!

luceos:
@mintunitish link?

drbyte:
I'm pleased to see @luceos taking this seriously. Adding `tenancy:key:rotate` or equivalent is a great idea.

I disagree with one of the comments above:
> The only reason to change keys is if you know it's been compromised.

This Laravel 5.6.30 patch due to a discovered risk is a perfect example of a reason to rotate keys, whether compromised or not.

While I realize many of us here aren't intending to submit our apps for PA DSS or PCI DSS certification, implementing their standards results in more secure applications. Secure coding practices should start at the beginning, not as an afterthought.

I also realize that APP_KEY in Laravel isn't a complete representation of a full on cryptographic key management strategy. 

Nevertheless, if one has never heard about this stuff, it's worth the 8 minutes to read the cheat-sheet summaries linked below.

OWASP points out that the PCI standards require at least annual rotation of cryptographic keys, plus also whenever a risk/threat of compromise is encountered:

> 3.6.4 Periodic cryptographic key changes
[...]

PR #657, opened 2018-11-07
Remove APP_KEY dependency for password generation

ghost:
This PR removes the `APP_KEY` from the password hashing.

The hash is generated through the following website properties:

- ID
- UUID
- Created at timestamp

All these parameters should never change in the application lifetime and provide a relative unique combination.

Changing the `APP_KEY` used to brick the application. With this PR you are able to rotate the `APP_KEY` as needed. Official recommendations for changing the `APP_KEY` in the past have also appeared: https://laravel.com/docs/5.6/upgrade#upgrade-5.6.30

Resolved issues: https://github.com/hyn/multi-tenant/issues/574

### DIFF
--- a/src/Generators/Database/DefaultPasswordGenerator.php
+++ b/src/Generators/Database/DefaultPasswordGenerator.php
@@ -37,9 +37,10 @@ class DefaultPasswordGenerator implements PasswordGenerator
     public function generate(Website $website) : string
     {
         return md5(sprintf(
-            '%s.%d',
-            $this->app['config']->get('app.key'),
-            $website->id
+            '%d.%s.%s',
+            $website->id,
+            $website->uuid,
+            $website->created_at
         ));
     }
 }


luceos:
We need a way to migrate existing installs. Otherwise merging this PR will disrupt production environments after blind upgrades. Hence being reluctant.

ghost:
@luceos I agree, that's why I think we should label this as a BC. Since changing the APP_KEY would have the same impact (and has probably already disrupted some environments), we are on a better path with this implementation.

Any ideas how a good migration process could be established? My suggestion would simply be an incompatible release without any migration.

twdnhfr:
Wouldn't it be better to just remove the need for a password for the tenant connection? As far as I can see the app_key is only used to have something random while creating but afterwards static. The tentant connection password isn't known to the "user" and is re-generated on each use. So you could create the mysql user without a password when connecting from localhost.

Or maybe as an alternative: using the .env for the tenant connection credentials just like the regular db connection

luceos:
> The tentant connection password isn't known to the "user" and is re-generated on each use.

The password is not re-generated and not using a password is very unsafe.

> Any ideas how a good migration process could be established? My suggestion would simply be an incompatible release without any migration.

Not having a migration plan is unacceptable. I think the best thing to do is write a migration command that updates users one at a time, this requires keeping knowledge about the old method as well.

twdnhfr:
So just being curious, where do you store the password for the tenant connection?

ghost:
> Not having a migration plan is unacceptable. I think the best thing to do is write a migration command that updates users one at a time, this requires keeping knowledge about the old method as well.

I understand, but using the `APP_KEY` is also unacceptable.

Do you have any ideas how that migration command could look like?

fletch3555:
> So just being curious, where do you store the password for the tenant connection?

We don't.  It's recalculated in code for each run, but not changed in the database.  E.g., if you know how to generate it (and have all the appropriate information), you would be able to log in manually at any point

luceos:
> Do you have any ideas how that migration command could look like?

Yes the command would resolve the Connection class of this package from IoC for each Tenant website. This connection class has a refactored method for the new configuration. A connection would be set up with the system connection of the tenant website (managed by property) by using the Connection class system() method, after which it updates the user password. This last step would require new methods on the existing Database Generator Drivers.

ghost:
> Yes the command would resolve the Connection class of this package from IoC for each Tenant website. This connection class has a refactored method for the new configuration. A connection would be set up with the system connection of the tenant website (managed by property) by using the Connection class system() method, after which it updates the user password. This last step would require new methods on the existing Database Generator Drivers.

Sounds good! You're more familiar with the codebase, can you add that command? :-)

luceos:
In hindsight I think generating this tenant password should have at least one element in the hasher string that is not available in the system database. As such it would be better to start using a custom TENANCY_KEY value in the future..

ghost:
> In hindsight I think generating this tenant password should have at least one element in the hasher string that is not available in the system database. As such it would be better to start using a custom TENANCY_KEY value in the future..

Not a good idea, this introduces the necessity of storing a permanent password. For what? Just to achieve the same result that the provided values of this PR also offer.

luceos:
All three values being used in this PR are part of the system database. Thus when the system database is breached, all tenants are exposed. I'd rather not introduce something that risky.

ghost:
@luceos so when the system database is breached, your solution provides more security? There is a big lack of realism here.

luceos:
> There is a big lack of realism here.

Care to explain @phpduck?

I am aware the system connection (per default) is supposed to have access to its tenants, but this is not a necessity. Lesser permissions to the system user (while stilling allowing it to create other databases and users) are still a feasibility. So I'd love to hear why my perspective is lacking realism.

ghost:
@luceos sure, these concerns are not within the scope of this package. Managing DB user permissions are done on another level, this makes your complete argument just invalid. And if your system db is breached, how can you guarantee that no access is provided to tenants with your key solution? That makes no difference to the other unique values. The system DB should be protected on a whole other level.

luceos:
The benefit of having one segment of the hashing outside of the system database is the ability to store it outside of harms way. Even though people with bad intent possibly gained access to the system database by checking the database.php or the .env aka the filesystem, this is not the only way to inject environment variables into an app. Another common approach is to use machine environment variables, eg when building docker images. And even though these might be read as well from the filesystem, when having gained access, other key stores like hashicorps vault aren't that easily accessible.

ghost:
@luceos that is way too specific, I prefer the more general approach here. However, if this PR doesn't suit the maintainers, it can be rejected/modified or whatever. IMHO the `APP_KEY` dependency is still a pain point which should be handled somehow.

Plytas:
I agree with both sides.
Tenancy should not depend on `APP_KEY` as it **needs** to be able and **should** rotate. Extra security with proposed `TENANCY_KEY` is invaluable. In the end if people don't want to use `TENANCY_KEY` they can just not specify it in `.env` and it would be an empty string by default.

luceos:
#739

PR #663, opened 2018-11-08
Fix MySQL privileges so that it is compatible with RDS

### LINKED ISSUE #660: `GRANT ALL` on Amazon RDS with MariaDB does not work
**Description**

Creating a new tenant with `auto-create-tenant-database-user = true` does not work on Amazon RDS with MariaDB.

*Per the discussion below it seems this problem only appears with MariaDB, not with MySQL.*

The problem is because the root MySQL user doesn't have `SUPER` privileges (see https://stackoverflow.com/a/44441811/245552).

> You can't grant SUPER privilege on RDS, so you can't use the ALL PRIVILEGES shortcut.
>
> Sorry, you must name the privileges you want to grant explicitly, even if it means listing every privilege except SUPER.

This piece of code fails because of the `GRANT ALL`:

https://github.com/hyn/multi-tenant/blob/37cf25b9810123643aa3f24f6bb9e8420ab13e42/src/Generators/Webserver/Database/Drivers/MariaDB.php#L52

Here is the official RDS documentation mentioning the privilege limits: https://aws.amazon.com/fr/premiumsupport/knowledge-center/duplicate-master-user-mysql/

```
mysql> SHOW GRANTS FOR 'master_user';

GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, RELOAD, PROCESS, REFERENCES, INDEX, ALTER, SHOW DATABASES, CREATE TEMPORARY TABLES, LOCK TABLES, EXECUTE, REPLICATION SLAVE, REPLICATION CLIENT, CREATE VIEW, SHOW VIEW, CREATE ROUTINE, ALTER ROUTINE, CREATE USER, EVENT, TRIGGER ON *.* TO 'master_user'@'%' IDENTIFIED BY PASSWORD '' WITH GRANT OPTION
```

---

__Actual behavior__

MySQL error.

__Expected behavior__

User is created and has the correct privileges.

---
__Information__

- hyn/multi-tenant version: 5
- laravel version: 5.6
- database driver and version: RDS, MariaDB
- webserver software and version: AWS Lambda / [Bref](https://github.com/mnapoli/bref)
- php version: 7.2

---

**Suggested solution**

I don't think the created users need all those privileges. We could fix this problem and increase security at the same time by listing the privileges actually needed.

**Do you know which privileges we should give the created users?**

In the meantime as a workaround we'll create the users manually with the specific privilege list (as explained [here](https://aws.amazon.com/fr/premiumsupport/knowledge-center/duplicate-master-user-mysql/)).

mnapoli:
Fix #660

### DIFF
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -49,7 +49,8 @@ class MariaDB implements DatabaseGenerator
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
-                return $connection->statement("GRANT ALL ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
+                $privileges = 'SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, RELOAD, PROCESS, REFERENCES, INDEX, ALTER, SHOW DATABASES, CREATE TEMPORARY TABLES, LOCK TABLES, EXECUTE, REPLICATION SLAVE, REPLICATION CLIENT, CREATE VIEW, SHOW VIEW, CREATE ROUTINE, ALTER ROUTINE, CREATE USER, EVENT, TRIGGER';
+                return $connection->statement("GRANT $privileges ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
             }
 
             return true;
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -49,7 +49,7 @@ class MariaDB implements DatabaseGenerator
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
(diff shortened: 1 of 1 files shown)


luceos:
I'm curious whether this will impact forward and backward compatibility with MySQL and MariaDB.. Tests for the current version seem to be AOK with this 🤔

mnapoli:
Here are the privileges that are "removed" with this change:

- SUPER
- FILE
- SHUTDOWN

Those don't seem like privileges you would want to give anyway right? (and that makes sense that they are not enabled on RDS)

The only exception might be `FILE`, this is the one that allows importing from CSV (LOAD DATA INFILE). But if you don't use it then it's fine.

luceos:
One advantage is that you can create a listener to add the necessary permissions after creation of the user anyway.

I'm still not too happy about this change, fearing for unexpected consequences.

I'll merge this as a feature for a new minor release.

LDP-Dries:
With regards to the CSV import, you could still technically use the system DB connection to import into a tenant DB. It's a little more hassle though.

mnapoli:
@luceos understood, I've also opened #665 if you prefer this: it adds a configuration option for the privilege list.

Feel free to choose this one or the other and close the PR you will not use :)

luceos:
Preferring the other one, thanks for the consideration. Great job!

PR #665, opened 2018-11-09
Allow to configure database privileges for the tenant DB user

### LINKED ISSUE #660: `GRANT ALL` on Amazon RDS with MariaDB does not work

mnapoli:
Fix #660, alternative to #663

Some databases like RDS do not allow granting all privileges. This new configuration option will let user provide a custom privilege list.

### DIFF
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -49,10 +49,7 @@ class MariaDB implements DatabaseGenerator
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
-                $privileges = config('tenancy.db.tenant-database-user-privileges', null);
-                if (!is_string($privileges)) {
-                    $privileges = 'ALL';
-                }
+                $privileges = config('tenancy.db.tenant-database-user-privileges', 'ALL');
                 return $connection->statement("GRANT $privileges ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
             }
 
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -49,7 +49,11 @@ class MariaDB implements DatabaseGenerator
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
-                return $connection->statement("GRANT ALL ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
+                $privileges = config('tenancy.db.tenant-database-user-privileges', null);
+                if (!is_string($privileges)) {
+                    $privileges = 'ALL';
+                }
+                return $connection->statement("GRANT $privileges ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
             }
 
             return true;
--- a/src/Generators/Webserver/Database/Drivers/MariaDB.php
+++ b/src/Generators/Webserver/Database/Drivers/MariaDB.php
@@ -49,7 +49,7 @@ class MariaDB implements DatabaseGenerator
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
-                $privileges = config('tenancy.db.tenant-database-user-privileges', 'ALL');
+                $privileges = config('tenancy.db.tenant-database-user-privileges', null) ?? 'ALL';
                 return $connection->statement("GRANT $privileges ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
             }
 
--- a/src/Generators/Webserver/Database/Drivers/PostgresSchema.php
+++ b/src/Generators/Webserver/Database/Drivers/PostgresSchema.php
@@ -29,7 +29,7 @@ class PostgresSchema extends PostgreSQL
 
     protected function grantPrivileges(IlluminateConnection $connection, array $config)
     {
-        $privileges = config('tenancy.db.tenant-database-user-privileges', 'ALL PRIVILEGES');
+        $privileges = config('tenancy.db.tenant-database-user-privileges', null) ?? 'ALL PRIVILEGES';
 
         return $connection->statement("GRANT $privileges ON SCHEMA \"{$config['schema']}\" TO \"{$config['username']}\"");
     }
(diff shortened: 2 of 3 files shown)


src/Generators/Webserver/Database/Drivers/MariaDB.php
```diff
@@ -49,7 +49,11 @@ public function created(Created $event, array $config, Connection $connection):
         };
         $grant = function ($connection) use ($config, $createUser) {
             if ($createUser) {
-                return $connection->statement("GRANT ALL ON `{$config['database']}`.* TO `{$config['username']}`@'{$config['host']}'");
+                $privileges = config('tenancy.db.tenant-database-user-privileges', null);
```
luceos: Instead of using `null` as second argument, use `ALL`, that makes the next line unnecessary.

src/Generators/Webserver/Database/Drivers/PostgresSchema.php
```diff
@@ -29,7 +29,12 @@ protected function createDatabase(IlluminateConnection $connection, array $confi
 
     protected function grantPrivileges(IlluminateConnection $connection, array $config)
     {
-        return $connection->statement("GRANT ALL PRIVILEGES ON SCHEMA \"{$config['schema']}\" TO \"{$config['username']}\"");
+        $privileges = config('tenancy.db.tenant-database-user-privileges', null);
```
luceos: Same as the other comment.

luceos (changes requested):
That's a great solution, would merge directly. Just left one comment for you to take a look at.

src/Generators/Webserver/Database/Drivers/MariaDB.php
mnapoli: Yes but I wanted to make sure that if the constant is defined to `null` (which is different than being undefined) then everything still works.

What do you think?

luceos: the config helper will return `null` if the configuration key cannot be found, so:

```php
config('some.foo', null); // null
```
is exactly the same as:
```php
config('some.foo'); // null
```

`undefined` does not exist in php

mnapoli: > if the configuration key cannot be found

yes but what if I have an environment variable set to an empty string in there? The variable will exist and will be defined to null so will it take the default?

I wanted to make sure there is no room for error. If you are sure that `config()` will take the default value *even if the config is defined explicitly to `null` then I'm fine changing my PR :)

luceos: Okay then rewrite to:

```php
$privileges = config('tenancy.db.tenant-database-user-privileges') ?? 'ALL';
```

It will do an isset, which also checks for empty string.

🤔 nope, both our versions won't work if the config var returns `''` an empty string. But I think that's not much of an issue, let's go for the first version, with the second argument of config being the default. @mnapoli

mnapoli: I would be careful with that too, try this:

```
FOO= php -r "var_dump(getenv('FOO') ?? 'ALL');"
```

This prints `string(0) ""`

So if a user do: `'tenant-database-user-privileges' => env('FOO'),` and `FOO=` in its `.env` then it will break I guess.

luceos: I guess so, but we're not being this strict anywhere else in the code.. It's your own responsibility to set up your environment and it would be pretty easy to debug as well..

mnapoli: OK I understand your call!

luceos: The tests are failing, let's go for the null coalescing operator instead and please try running tests locally before pushing.
```php
$privileges = config('tenancy.db.tenant-database-user-privileges') ?? 'ALL';
```

mnapoli: Oh right sorry.

luceos (approved):

luceos:
It's all green!! Thanks for your patience.

HpiGuillaume:
@mnapoli @luceos Thanks a lot !!!

PR #671, opened 2018-11-22
Check install status only once by request

bastien-phi:
This PR changes `$installed` to static in class `Environment`.

During application bootstrap, `Environment` is constructed several times and `$isInstalled` function is called each time, causing several database queries.

Making `$installed` static fixes this by sharing install status between each instance of `Environment`.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -38,7 +38,7 @@ class Environment
     /**
      * @var bool
      */
-    protected static $installed;
+    protected $installed;
 
     public function __construct(Application $app)
     {
@@ -70,7 +70,7 @@ class Environment
             }
         };
 
-        return self::$installed ?? self::$installed = $isInstalled();
+        return $this->installed ?? $this->installed = $isInstalled();
     }
 
     public function identifyHostname()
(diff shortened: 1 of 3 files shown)


luceos:
Even though I understand this would be a fix. The underlying problem lies in this:

> During application bootstrap, Environment is constructed several times and $isInstalled function is called each time, causing several database queries.

The Environment should be bound into the Service Container as a singleton and never be instantiated multiple times. Have you tried checking why this happens?

bastien-phi:
Problem comes from the fact the singleton is registered after other providers are booted (for exemple `HostnameProvider`).

Actually, `Environment` is constructed 7 times.

I found a solution for that, just need to dig it a bit, test and make it clean.

Do you want another pull request for that or can I just modify this one ?

bkintanar:
@bastien-phi modifying this one is fine. Thanks!

bastien-phi:
Commit https://github.com/hyn/multi-tenant/pull/671/commits/1ecb94335372b4b5f46fe29342cd903a25c6f913 does fix multi instanciation of `Environment`. Just need to declare early the singleton, it will be constructed when needed and only once.

luceos:
Thanks!

Issue #566, opened 2018-07-30
Package events not firing / not being captured


DennisOng:
__Description__

Seems like this package's events are not firing for me, or rather, I'm unsuccessful in capturing those events.

---

__Actual behavior__

Events not captured

__Expected behavior__

Events captured

---
__Information__

- hyn/multi-tenant version: 5.2.6
- laravel version: 5.6.29
- database driver and version: mysql
- php version: 7.2

---

__tenancy.php config__

```php
    'models' => [
        /**
         * Specify different models to be used for the global, system database
         * connection. These are also used in their relationships. Models
         * used have to implement their respective contracts and
         * either extend the SystemModel or use the trait
         * UsesSystemConnection.
         */

        // Must implement \Hyn\Tenancy\Contracts\Hostname
        'hostname' => \Hyn\Tenancy\Models\Hostname::class,

        // Must implement \Hyn\Tenancy\Contracts\Website
        'website' => \Hyn\Tenancy\Models\Website::class
    ],
    'website' => [
        /**
         * Each website has a short random hash that identifies this entity
         * to the application. By default this id is randomized and fully
         * auto-generated. In case you want to force your own logic for
         * when you need to have a better overview of the complete
         * tenant folder structure, disable this and implement
         * your own id generation logic.
         */
        'disable-random-id' => false,

        /**
         * The random Id generator is responsible for creating the hash as mentioned
         * above. You can override what generator to use by modifying this value
         * in the configuration.
         *
         * @warn This won't work if disable-random-id is true.
         */
        'random-id-generator' => Hyn\Tenancy\Generators\Uuid\ShaGenerator::class,

        /**
         * Enable this flag in case you're using a driver that does not support
         * database username or database name with a length of more than 32 characters.
         *
         * This should be enabled for MySQL, but not for MariaDB and PostgreSQL.
         */
        'uuid-limit-length-to-32' => env('LIMIT_UUID_LENGTH_32', true),

        /**
         * Specify the disk you configured in the filesystems.php file where to store
         * the tenant specific files, including media, packages, routes and other
         * files for this particular website.
         *
         * @info If not set, will revert to the default filesystem.
         * @info If set to false will disable all tenant specific filesystem auto magic
         *       like the config, vendor overrides.
         */
        'disk' => 's3',

        /**
         * Automatically generate a tenant directory based on the random id of the
         * website. Uses the above disk to store files to override system-wide
         * files.
         *
         * @info set to false to disable.
         */
[...]

PR #673, opened 2018-11-22
Add support for division mode to separate database with shared credentials as system

nevertry:
This PR adds ability to use new tenant division mode: `DIVISION_MODE_SEPARATE_SHARED_ACCESS` / `shared`

Use case:

I usually manage / monitor database using DBeaver, that shows all databases regardless of privileges for the user I was logged to.

However, in RDS setup, `rds_superuser` role  is not really a superuser, it discourages user to switch access to another database / `use other_database` and doing query, instead must create new/edit connection and login with different credentials.
Which is still not allowed after #665 addition.

In my case, tenant's user will not be given access to database directly.

Thanks.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -45,6 +45,11 @@ class Connection
     const DIVISION_MODE_SEPARATE_PREFIX = 'prefix';
 
     /**
+     * Allows multiple database but with same user access
+     */
+    const DIVISION_MODE_SEPARATE_SHARED_ACCESS = 'shared';
+
+    /**
      * Allows division by schema. Postges only.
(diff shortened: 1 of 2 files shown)


luceos:
I think listening to the Configured event and overriding username/password back to the system connection configuration would be more than enough. Taking this event into consideration, your implementation adds nothing one cannot do currently. I also think adding another division mode is overly complex compared to using an Event listener.

nevertry:
Okay, thank you sir.

I have issue registering event listener, but solved with this: https://github.com/hyn/multi-tenant/issues/566#issuecomment-415670494

PR #683, opened 2018-12-07
Respect vendor assets overrides

bastien-phi:
This PR allows to override vendor views and vendor translations at tenant level when enabled : 
- For translations
```php
'trans' => [
    'enabled' => true,
    'override-global' => true,
    // ...
]
```
- For views 
```php
'views' => [
     'enabled' => true,
    'namespace' => null,
     'override-global' => true,
],
```

It also fixes override of namespaced vendor translations in `resources/lang/vendor` (for example, I was having `'nova::validation.attached'` in nova when doing `trans('nova::validation.attached')`.


luceos:
Nice job!

Issue #576, opened 2018-08-11
Ability to make 3rd party package jobs tenant aware


Plytas:
__Description of your feature__

I'm experimenting with https://github.com/spatie/laravel-medialibrary and they provide an option to generate image conversions through a queue worker. Unfortunately, there is no way to make their jobs tenant aware.

luceos:
See https://github.com/spatie/laravel-medialibrary/pull/377#issuecomment-412487156

Plytas:
This can be closed as my specific issue will be solved. Though I'm wondering if it's actually possible to somehow intercept jobs and add tenant information to them.

luceos:
I'm leaving this open to increase integration with other packages. I'm not sure it's possible, but it's worth a shot.

luceos:
This should be fixed with #711

ArlonAntonius:
We just merged and tagged a new version that includes the feature discussed in this PR:
- Job Auto Tenancy

Also: Spatie's MediaLibrary package allows for overriding their jobs by your own.

If this doesn't actually solve your issue, you can always contact us through our channel(s):
- [Discord](https://laravel-tenancy.com/chat)

PR #711, opened 2019-02-07
Job auto tenancy

vpratfr:
Implement automatic website ID serialization as per M. Said article on https://divinglaravel.com/preparing-jobs-for-queue

### DIFF
--- a/tests/unit-tests/Providers/TenancyProviderTest.php
+++ b/tests/unit-tests/Providers/TenancyProviderTest.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of the hyn/multi-tenant package.
+ *
+ * (c) Daniël Klabbers <daniel@klabbers.email>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @see https://laravel-tenancy.com
+ * @see https://github.com/hyn/multi-tenant
+ */
+ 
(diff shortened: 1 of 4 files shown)


luceos:
fixes #576

tests/unit-tests/Providers/TenancyProviderTest.php
```diff
@@ -71,7 +72,7 @@ public function current_website_id_is_included_in_notification_job_payload()
     {
         Event::fake();
 
-        $user = $this->userStory()->create();
```
luceos: I was so confused about this one @vpratfr, I almost thought I missed a new feature in Laravel 😊

luceos:
One little request. Can you move the logic inside the TenancyProvider to a QueueProvider in the tenants subdirectory, so that it matches the project structure?

Have you been able to fix the tests?

ArlonAntonius:
One thing we might want to keep in mind on this as well:
What if someone sets the website_id themselves, we shouldn't overwrite it in that case as it would cause things to run on a different tenant than they expect.

vpratfr:
Won't be able to address changes soon, sorry. I have a very busy week coming plus winter holidays

vpratfr:
:thumb:

ArlonAntonius:
Just realized that this TenancyProvider uses the Website model directly, the one we define in the package.
We should refrain from using this as we provide a way for others to use their own models.

vpratfr:
True. The config model shall be used indeed.

PR #712, opened 2019-02-07
Adding regex validation to HostnameValidator

ArlonAntonius:
Original PR by: @fletch3555 
All credits go to him.

Simply test case created based on his info, Github is acting weird so that's why this is a new PR.

### DIFF
--- a/src/Validators/HostnameValidator.php
+++ b/src/Validators/HostnameValidator.php
@@ -28,7 +28,7 @@ class HostnameValidator extends Validator
 
     protected $update = [
         'id' => ['required', 'integer'],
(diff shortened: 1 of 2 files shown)


luceos:
The only concern I have (other than the tests failing, not sure why yet) is that the hostname validator might now be too strict for local development. Can we include local domains as well in the test, eg `localhost` for simplicity sake?

Tests are failing on the new code.

ArlonAntonius:
@luceos 
As per your message here and the quick discussion on Discord, I'll provide a more detailed answer here for documentation purposes as well.

Local Development
We should support Local Development as much as we can, however, at the end of the day, this is meant for production and not keeping things local. When checking the regex (on [Regex101](https://regex101.com/) we can see it matches exactly the domains as @fletch3555 stated.
I added a few local development domains based on the link you provided on discord and common sense.
image

"localhost" is not matching and I think we should keep it that way. If people would develop a multi-tenant application on a local machine, it would still use subdomains and those will match using the new HostnameValidator.

If you want to actually localhost as matching as well, you can try something like this as the regex:
`^((?:(?:\*|(?!-)(?:xn--)?[a-zA-Z0-9][a-zA-Z0-9-_]{0,61}[a-zA-Z0-9]{0,1})\.)(?:(?!-)(?:xn--)?[a-zA-Z0-9][a-zA-Z0-9-_]{0,61}[a-zA-Z0-9]{0,1}\.)*(?!-)(?:xn--)?(?:[a-zA-Z0-9\-]{1,50}|[a-zA-Z0-9-]{1,30}\.[a-zA-Z]{2,})|localhost)$`

Hardcoded localhost as simply allowing words would result in also matching "bread" or things like that and I think that's quite some weird behavior.

PR #731, opened 2019-02-26
Laravel 5.8 Compatibility

ArlonAntonius:
As per Laravel 5.8 the Defer variable in Service Providers is [deprecated](https://github.com/laravel/framework/pull/27067). This PR simply fixes the deferrable provider per their requirements, would break compatibility with 5.7 though.


ArlonAntonius:
Few things that still need to be done:
- Require Laravel / Helpers (only for dev, we use array_forget in InteractsWithTenancy)
- Use Array facade inside test, instead of function

Miguel-Serejo:
To maintain compatibility with 5.7, you could leave the `$defer` property in place and just implement the new interface.

ArlonAntonius:
@36864 The interface is not available in L5.7, meaning you'll just get an error saying the Interface does not exist?

Miguel-Serejo:
And that's what happens when I try to review code before my morning coffee. Sorry.

It's worth pointing out that this will also break compatibility with 5.5, which is the current LTS branch. Maybe we can delay that until the next version, seeing as this package claims to be compatible with latest stable and LTS?

ArlonAntonius:
@36864 
#### Delay the Release
At this moment we have not planned a new release of the package yet as there are still errors to take care of. I don't see the fixing of that happening any time soon as we have close to no idea where the exact error is coming from and it's something that still needs to be investigated. This is just a PR that makes the package 5.8 compatible and future-proof (by already removing deprecated things).
What we could do is create a separate PR that changes only the actual removed things to prevent breaking older versions as much as possible.

#### Breaking Compatibility
The current code that is in the dev branch is already breaking 5.5 as functionality that got introduced in #711 uses functions that have been added since L5.7.

PS: This is just my opinion / view on things

Miguel-Serejo:
Fair point about the compatibility, I only brought it up because the readme claims compatibility with latest LTS. 

I'm just gonna go back to silently appreciating your work now.

luceos:
Seems we need to clarify the readme. The new strategy is to stay up to par with the latest release. 

I'm going to invest some time again this week to assist @ArlonAntonius in getting a release out.

PR #734, opened 2019-02-27
wip readying up for 5.4.0 and laravel 5.8

luceos:


ArlonAntonius:
If we decide to take this approach of updating to PHPUnit8, we still need to bump the PHP version in the composer, not doing that now as the decision is up to @luceos

itguy614:
Will 5.3 of tenancy be updated for L5.8 support or will that only be included in 5.4?  Perhaps it wouldn't hurt to align the version numbers once 5.4 is released.

ArlonAntonius:
@itguy614 
Why would we support 5.3 with Laravel 5.8?
What would be the real purpose of that if I may ask?

Don't think we need to align the version numbers as it would kind of limit us as well in how much we can update / change.

ArlonAntonius:
(Wrong button, sorry)

src/Abstracts/WebsiteEvent.php
```diff
@@ -23,8 +24,14 @@ abstract class WebsiteEvent extends AbstractEvent
      */
     public $website;
 
-    public function __construct(Website &$website)
+    /**
+     * @var Hostname
+     */
+    public $hostname;
+
+    public function __construct(Website &$website, Hostname $hostname = null)
```
morloderex, line 32: Hostname should be

```suggestion
    public function __construct(Website &$website, ?Hostname $hostname = null)
```

luceos: Optional function arguments aren't using the question mark, but a default value `null`, check the php documentation for more information:

http://php.net/manual/en/migration71.new-features.php

image

PR #737, opened 2019-03-01
ability to enable different webserver based on environment

roelgonzalez:
I have come across situation where I had to work with two different webservers temporarily, although this may not be an ideal situation to have it does help to be able to configure this.

### DIFF
--- a/assets/configs/webserver.php
+++ b/assets/configs/webserver.php
@@ -24,7 +24,7 @@ return [
         /**
          * Whether the integration with Apache2 is currently active.
          */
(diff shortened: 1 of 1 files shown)


Plytas:
As these are config file changes only, they can be easily implemented in your own application. I don't think this should be merged into the package.

roelgonzalez:
True, it is easily changeable based on your personal requirements. Feel free to close.

drbyte:
@roelgonzalez since you opened the issue/PR you can also close it, using the Close button below.

PR #738, opened 2019-03-01
abort to default hostname upon non-identification without required tenant

roelgonzalez:
The way the package works now is that when a tenant is not identified it will do the following:

1. Check to see if you want to abort upon no identification. If so then simply 404.
2. If you have a fallback, go to that tenant, however you are required to have that tenant setup.

I thought it would be nice to have a full back location without actually needing to have a tenant. Maybe that would be considered the landing page for the SaaS.

Let me know what you guys think.

### DIFF
--- a/assets/configs/tenancy.php
+++ b/assets/configs/tenancy.php
@@ -121,6 +121,16 @@ return [
          * @warn this must be a FQDN, these have no protocol or path!
          */
         'default' => env('TENANCY_DEFAULT_HOSTNAME'),
+        
+        /**
+         * If you want the multi tenant application to fall back to a fallback
+         * url, without requiring it to be an actual tenant in the system,
+         * complete in detail the default fallback url with protocol.
+         * 
+         * e.g. https://example.org
+         */
+        'fallback-url' => env('TENANCY_FALLBACK_URL'),
+        
         /**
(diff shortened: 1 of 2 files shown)


roelgonzalez:
I have noticed a flaw, this should not be merged (yet)

roelgonzalez:
I have made another commit to fix an issue where you couldn't go to anywhere but the base url.

luceos:
You are re-using a configuration item used for something else entirely. Without a (new) flag/configuration item your suggested solution will cause misbehaviour on existing installations, that's a huge no. I'm okay with this solution, but only if you add another configuration item for this behaviour.

roelgonzalez:
The way it is right now will work just fine as far as I know, however I've been wondering if we need to make this compatible with sub-domains that are also not necessarily a tenant.

However I'm starting to lean towards that not being necessary to add to the package, as one could simply make a virtual host nginx / apache configuration that will catch that non-tenant specific situation.

Let me know thoughts.

luceos:
I think we're not going to add this to the package after all. Reason for this is simple, the HostnameActions is meant as an inspiration. There's absolutely no argument to keep using that Middleware in case you want something entirely different. Let's stick to this approach, the tenancy.php configuration already allows configuring what middleware is active.

Issue #767, opened 2019-03-27
MediaController - Images not working


xalunda:
Good afternoon,

I'm trying to implement the `media` route for my tenants.
Here are my configurations:

---

`routes/tenants.php`:
```
// MEDIA
Route::get('/media/{path}', '\Hyn\Tenancy\Controllers\MediaController')
->where('path', '.+')
->name('tenant.media');
```

---

`storage`:
```
/storage/app/public/tenancy/d3442a527d7b4b08a8053755b87151f1/media/files/
```
where `d3442a527d7b4b08a8053755b87151f1` is my tenant's uuid

---
__My Issue__:

The `/media` route doesn't seem to work with images.
I tested `.png`, `.jpeg`, `.jpg` and `.svg`.
However, it works fine with `.pdf` files.

---

__Image Test__:
I have an image `example.jpeg`

Here are the response headers received from `https://tenant.domain.local/media/files/example.jpeg`:

```
Cache-Control: no-cache, private
Connection: Keep-Alive
Content-Type: image/jpeg
Date: Wed, 27 Mar 2019 16:34:42 GMT
Keep-Alive: timeout=5, max=100
Server: Apache
Transfer-Encoding: chunked
X-Powered-By: PHP/7.2.1
```
This method sends a broken image, whatever the image is:
![image](https://user-images.githubusercontent.com/2316221/55094810-041cb580-50af-11e9-95fb-b58e735522c9.png)

Here are the response headers accessing the file directly `https://tenant.domain.local/storage/tenancy/d3442a527d7b4b08a8053755b87151f1/media/files/example.jpeg`:

```
Accept-Ranges: bytes
Connection: Keep-Alive
Content-Length: 4960688
Content-Type: image/jpeg
Date: Wed, 27 Mar 2019 16:36:06 GMT
ETag: "2018c7a9f-4bb1b0-5572d22e2ff80"
Keep-Alive: timeout=5, max=100
Last-Modified: Sun, 20 Aug 2017 10:50:54 GMT
Server: Apache
Vary: User-Agent
```
This sends correctly the image.

I also tried to modify `MediaController` so that it includes the missing headers. No luck.
```
public function __invoke(string $path)
{
    $path = "media/$path";
    if ($this->directory->exists($path)) {
        $file = $this->directory->get($path);
        return response($file)
            ->header('Content-Type', Storage::disk('tenant')->mimeType($path))
            ->header('Accept-Ranges', 'bytes')
[...]

PR #769, opened 2019-03-28
#767 - MediaController - Images not working

xalunda:
Laravel framework might be introducing whitespace which might be ruining the `header()` function.

Use `ob_end_clean()` before our first `header()` call to remove any extra whitespace.

Closes [#767 - MediaController - Images not working](https://github.com/hyn/multi-tenant/issues/767)


luceos:
As discussed in issue, closing this because this proposal is too drastic and doesn't fix possible underlying issue.

PR #788, opened 2019-04-20
Override Default Model Connection

joshjacks:
Allows overriding the connection set by the traits ```UsesTenantConnection``` and ```UsesSystemConnection```.

### Use Case

We have an ```Address``` model used with several other models in our tenant dbs. Our customers, stored in the system db also have an address. We ideally do not want to create a second identical model.

Tried using ```Address::on('...')->...``` as suggested [here](https://github.com/tenancy/multi-tenant/issues/241#issuecomment-327406281). This doesn't currently work if the model has either of the above traits.

### DIFF
--- a/src/Traits/UsesSystemConnection.php
+++ b/src/Traits/UsesSystemConnection.php
@@ -20,6 +20,6 @@ trait UsesSystemConnection
 {
     public function getConnectionName()
     {
-        return app(Connection::class)->systemName();
+        return $this->connection ?: app(Connection::class)->systemName();
     }
 }
--- a/src/Traits/UsesTenantConnection.php
+++ b/src/Traits/UsesTenantConnection.php
@@ -20,6 +20,6 @@ trait UsesTenantConnection
 {
     public function getConnectionName()
     {
-        return app(Connection::class)->tenantName();
+        return $this->connection ?: app(Connection::class)->tenantName();
     }
 }
--- a/tests/unit-tests/Models/SystemModelTest.php
+++ b/tests/unit-tests/Models/SystemModelTest.php
@@ -29,4 +29,20 @@ class SystemModelTest extends Test
         $this->assertEquals($model->getConnectionName(), $this->connection->systemName());
         $this->assertEquals($model->getConnection(), $this->connection->system());
     }
+
+    /**
+     * @test
+     */
+    public function can_override_connection()
+    {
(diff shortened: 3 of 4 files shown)


Plytas:
There's a `TenantAwareConnection` trait, have you tried using that?

joshjacks:
Hey @Plytas, thanks for the suggestion. Looking at this trait, it looks like it will choose the system connection only if the tenant connection is not set up? The above allows overriding even after the tenant connection is setup.

E.g. after this PR, we could do something like this

```php
Address::create(['postcode' => 'XX0YY7']);
Address::on('system')->create(['postcode' => 'ZZ0PP7']);
```

I don't think the ```TenantAwareConnection``` allows this? Correct me if I'm wrong 😄

luceos:
I'm not entirely against this, but I'm also not seeing any benefit. You can easily create your own trait for your code base. Not everyone has the exact needs you have.

joshjacks:
Hi @luceos, I agree. Of course, it is up to you guys what you think is best for the package. We did end up using our own trait but think that the proposed change felt more in line with what would be expected based on the current default Laravel behaviour. 

I'm sure they exist, but we couldn't think of any use cases that would require preventing overriding like this.

Appreciate the feedback 😄

azeos:
I think we are facing a similar problem.

We are using Laravel notifications to get the likes on posts (just like Facebook). The notifications table and the users table are in the system database. A user could have access to different tenants, so no matter in which tenant he is currently on, he will get notifications of all the websites.

Each notification shows the post's title that was liked. We store in each notification the "website_id" and the "post_id" (the posts table is in each tenant).

Now, the `Post` model uses the `UsesTenantConnection` trait. The idea was to get the posts doing something like this:

```PHP
foreach ($notifications as $notification) {
    $post = Post::on($notification->website->uuid)->find($notification->post_id);
}
```
So, if the user is currently on **Tenant A** but we need to get the `Post` from **Tenant B** we use the `on($website->uuid)`, but it seems that gets ignored.

I'll try this trait. Will you consider adding this functionality?  
I think that at least should be clear that you can't use `on()` if you are using the `UsesTenantConnection` trait.

Thanks!

**EDIT:**  
Well, didn't think about it, it won't work because the UUID database doesn't exists in the database config.

> Database [01f5a44d8e2d4eb389b48ab2922c20ba] not configured.

PR #805, opened 2019-05-22
Fix parsing of url scheme in UpdateAppUrl listener

JaZo:
The parenthesis was at the wrong place making PHP_URL_SCHEME the fallback for the app.url config variable.


luceos:
🤦‍♂ gosh, that's embarrassing, thanks!

PR #821, opened 2019-07-24
Use entire request URI instead of only path when forcing HTTPS

JaZo:
Query parameters are stripped from the (secure) URL when using `$request->path()`. I changed it to `$request->getRequestUri()` to make sure query parameters are included.

**Before**
`http://example.com/page?foo=bar -> https://example.com/page`

**After**
`http://example.com/page?foo=bar -> https://example.com/page?foo=bar`

### DIFF
--- a/src/Middleware/HostnameActions.php
+++ b/src/Middleware/HostnameActions.php
@@ -95,7 +95,7 @@ class HostnameActions
     {
         $this->emitEvent(new Secured($hostname));
 
-        return $this->redirect->secure($request->path());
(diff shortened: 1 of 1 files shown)


fletch3555 (approved):
Seems reasonable to me.

drbyte:

JaZo:
> Updating this PR by merging recent changes from Master should let the tests pass.

I rebased onto 5.x!

luceos:
Thank you! Sorry for the huge delay 😊

PR #822, opened 2019-07-24
Change apt mysql client

ArlonAntonius:


fletch3555 (approved):

fletch3555:
Ugh..... current failure is due to mariadb 10.4.*.... https://mariadb.org/authentication-in-mariadb-10-4/

Apparently they changed to an "auth plugin" model, so the default "user" table is now a view, which is read-only (thus the error).  Might want to just peg the version to something a bit older for now

drbyte:

ArlonAntonius:
@drbyte This was not needed anymore, thanks for the bump 😄

PR #830, opened 2019-08-09
Set a limit on explode() in RunCommand

stancl:
I was implementing this feature for my package and took some inspiration from how you handle passing options/args to the command. Thought I'd contribute back.

If you use a limit on `explode()`, it will let you have a `=` in the value, rather than just cutting `--argument="subject=foo=bar"` into `'--subject' => 'foo'`.


fletch3555 (approved):
Sounds good to me

bkintanar (approved):

drbyte:

luceos:
Thanks, makes sense! 👍

PR #837, opened 2019-08-30
Force the default connection to 'tenant' on tenancy:run with --tenant

mertyildiran:
This pull request forces the default connection to 'tenant' for Artisan commands called with `php artisan tenancy:run` when `--tenant` option is present as it's suggested in [the documentation](https://tenancy.dev/docs/hyn/5.4/connections#forcing-the-connection). 

Since packages like `cartalyst/sentry` might contain models and can be directly called from Artisan commands for tenants, it seems like a necessity to force the connection with `config(['database.default' => 'tenant']);` in here.

### DIFF
--- a/src/Commands/RunCommand.php
+++ b/src/Commands/RunCommand.php
@@ -29,7 +29,7 @@ class RunCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'tenancy:run {run : The artisan command to run for the tenants} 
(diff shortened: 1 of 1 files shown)


luceos:
Let's assume you use the system connection as default. Merging this would impact any model supposed to work on the system connection and thus break every app.

I personally think if you really need this, you can override/extend the command and do so. The default behavior we have right now isn't really needing this forced connection.

PR #847, opened 2019-10-01
Move registration of FlushHostnameCache to TenancyServiceProvider::boot 

JaZo:
## Description

I moved the registration of `FlushHostnameCache` to `TenancyServiceProvider::boot`.

## Motivation and context

The event dispatcher, which is required for model events, is only available in the boot method of a service provider. This event dispatcher is setup in `\Illuminate\Database\DatabaseServiceProvider::boot`. Registering an event observer earlier will cause the model to be booted too early when it has no event dispatcher. A model will only boot once, so if the consumer has custom logic in his model boot method that requires an event dispatcher, that functionality is also silently broken (that's how I discovered this issue).

### TL;DR
`FlushHostnameCache` is not activated.

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/src/Providers/TenancyProvider.php
+++ b/src/Providers/TenancyProvider.php
@@ -60,6 +60,8 @@ class TenancyProvider extends ServiceProvider
     public function boot()
     {
         $this->bootCommands();
+
+        $this->bootObservers();
     }
 
     protected function registerModels()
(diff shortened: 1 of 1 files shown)


luceos:
Thanks for the great clarification. Totally makes sense!

PR #859, opened 2019-10-30
access modifier of drivers method from private to protected for easy override on DatabaseDriverProvider

awebartisan:
In order to override the database driver classes provided by the package, one could extend the DatabaseDriverProvider and override the `drivers` method. At the moment, this method is "private", therefore unable to override. This pull request changes its access modifier from 'private' to 'protected'.


luceos:
Always protected, not private, totally agree 👍

PR #914, opened 2020-03-25
bump uuid

creati0ns:
uuid gives conflicts with paragonie, when using roave/security-advisories during installation. Indication that there is a vulnerable dependant package.


bkintanar:
Sorry didn't saw this earlier. Already merged a fix #918. Closing this one.

PR #932, opened 2020-04-19
Allow to configure the database user host for MySQL and MariaDB

mfcanovas:


ArlonAntonius:
This is possible by listening to the `Creating` event and adjust the `host` in the `config`.

For that reason, I'd prefer not adding another config value to the already huge config.

mfcanovas:
I understand but keep in mind that current behaviour just doesn't work for proxy connections like using [Google Cloud SQL behind a proxy](https://cloud.google.com/sql/docs/mysql/sql-proxy).

Maybe it's better not to assume that user host can be inferred from database connection host and leave % as default value which won't cause problems to anyone and could be overriden by listening to the Creating event if you need more granular control about database user permissions.

marnickmenting:
I was struggling with this too. For reference, here is some example code that works for me in Laravel 9:

app/Listeners/TenantDatabaseCreating.php:
```
<?php

namespace App\Listeners;

use Hyn\Tenancy\Events\Database\Creating;

class TenantDatabaseCreating
{
    /**
     * Create the event listener.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Handle the event.
     *
     * @param  object  $event
     * @return void
     */
    public function handle(Creating $event)
    {
        // Change the host where the MySQL user should be given access to
        // You could replace this with a config() call if you want
        $event->config['host'] = '%';
    }
}
```

In EventServiceProvider.php:
```
use Hyn\Tenancy\Events\Database\Creating;
use App\Listeners\TenantDatabaseCreating;

...
protected $listen = [
...
Creating::class => [
            TenantDatabaseCreating::class
        ]
];
```

PR #940, opened 2020-05-05
Add new conditions to enable maintenance mode

Abbotton:
First of all, **thank you very much** for the development team of this package.

Now I find that this package has a problem in judging whether to enable `maintenance mode.` Now suppose I create a tenant account and set the tenant expiration time to 2021-05-05 (that is, the value of the `under_maintenance_since` field is 2021-05- 05), after the tenant account is successfully created, access the fqdn corresponding to the tenant, it shows **503 maintenance mode**, I think there is no reason.

Now if I want to solve my problem, I need to set the `under_maintenance_since value` to `null`, and add an expiration time field `expire_at` , and finally set a scheduled task to be executed once a day or every hour. Once the tenant account is found to be expired, set The value of `under_maintenance_since` is the value corresponding to `expire_at`.

But this is not the best solution, the best solution is to modify this package, so I did so.

Perhaps my understanding of the architecture of this package is incorrect, and I hope someone can answer this question for me.

**Thanks again**.

### DIFF
--- a/src/Middleware/HostnameActions.php
+++ b/src/Middleware/HostnameActions.php
@@ -57,7 +57,7 @@ class HostnameActions
             : null;
 
         if ($hostname != null) {
-            if ($hostname->under_maintenance_since) {
+            if ($hostname->under_maintenance_since && $hostname->under_maintenance_since <= now()) {
                 return $this->maintenance($hostname);
             }
 


bkintanar:
Thank you for your contribution, looks to be a solid code.

Issue #905, opened 2020-03-09
url() helper ignoring port when update-app-url setting is true


Bibendus83:
I have my dev environment working on a port different from 80/443.
If I set `update-app-url` in `tenancy.php` to true all my `url()` helper calls return the url without including the port.
Checking `Hyn\Tenancy\Listeners\URL\UpdateAppUrl` force() method I noticed that the url rewrite does not include the port
`$url = sprintf('%s://%s', $scheme, $hostname->fqdn);`

Shouldn't the port be included?

ArlonAntonius:
Hi there,

This is a known issue and is something that is simply not supported in the package at this moment. We really work with URL's and not with ports.

webard:
@ArlonAntonius but the port is part of URL...

bkintanar:
@webard FQDN != URL
![image](https://user-images.githubusercontent.com/685928/76325424-3d7f1d00-6322-11ea-89a5-28507f5c6d45.png)

Bibendus83:
but we are talking about the `url()` helper so it should contain a full URL not only the FQDN

ArlonAntonius:
Hi @Bibendus83,

Tenant's have an FQDN that is used for the URL and is excluding a port. This is something that is known and it's simply not something that we support at this moment.

Bibendus83:
It's not a question of supporting it or not, using this setting causes an error on the default laravel helper. 
Shouldn't it be fixed?

ArlonAntonius:
What exactly is the error?
Could you provide a stacktrace?
What is the fqdn set for the tenant?

fletch3555:
I have a few thoughts on this.  I definitely understand where @Bibendus83 is coming from.  It's not necessarily that the functionality is "wrong" as much as it's "counter-intuitive"

In general, modern websites/webapps really _shouldn't_ live on any port other than 80/443.  If you need to (for the sake of supporting multiple sites/apps on the same host, then that can easily be done with domain-based vhosts in the webserver(apache/nginx/etc) instead of by port.  The backend(php/laravel) layer of your application shouldn't care what port it's running at.  The exception to this is if you're attempting to distinguish your tenants by port number, but that doesn't appear to fit your requirements (and is honestly a pretty terrible idea in general).

[...]

PR #951, opened 2020-07-06
Addedd missing port to url generation when update-app-url is set to true

Bibendus83:
When a different port from from 80 or 443 is used in the url (generally on dev environments) the port was not added to the generated urls.
This should be a safe fix considering that the url generator behaviour is not altered when ports 80 or 443 are used.
Solving issue on ticket #905


ArlonAntonius:
@bkintanar Can you check if this is something we really should do? It would improve port compatibility but feels like something bloated to me. Upto you what happens with this.

eliurkis:
This is helpful in environments like CircleCI where you can't use 80 or 443 for the testing.

👍 💪

PR #955, opened 2020-07-28
Fix Schema Division Handling

mikebronner:
This PR aims to fix some issues with with schema division:

- only create the schema if it doesn't already exist.
- don't drop privileges for schemas.
- drop schema using cascade in order to clear out its contents.


src/Generators/Webserver/Database/Drivers/PostgresSchema.php
```diff
@@ -54,6 +54,6 @@ public function updated(Updated $event, array $config, Connection $connection):
 
     protected function dropDatabase(IlluminateConnection $connection, array $config)
     {
-        return $connection->statement("DROP SCHEMA IF EXISTS \"{$config['schema']}\"");
+        return $connection->statement("DROP SCHEMA IF EXISTS \"{$config['schema']}\" CASCADE");
```
ArlonAntonius, line 57: Using `CASCADE` is quite a big danger it, as far as my PSQL knowledge goes, this would delete everything in the schema. I think it would be better to use `RESTRICT` **if** we should change this at all. I think `RESTRICT` is even the default.
If we change this, we could risk people ending up deleting entire schema's while that's something that they would not want.

src/Generators/Webserver/Database/Drivers/PostgreSQL.php
```diff
@@ -129,6 +129,10 @@ protected function flushConnection(IlluminateConnection $connection, array $conf
 
     protected function dropPriviliges(IlluminateConnection $connection, array $config)
     {
+        if (config("tenancy.db.tenant-division-mode") === Connection::DIVISION_MODE_SEPARATE_SCHEMA) {
+            return true;
+        }
```
ArlonAntonius, line 134: Can you give a little insight into why we should not do this?

src/Generators/Webserver/Database/Drivers/PostgreSQL.php
```diff
@@ -143,8 +147,9 @@ protected function dropDatabase(IlluminateConnection $connection, array $config)
 
     protected function dropUser(IlluminateConnection $connection, array $config)
     {
-        if (config('tenancy.db.auto-delete-tenant-database-user') && $this->userExists($connection,
-                $config['username'])) {
+        if (config('tenancy.db.auto-delete-tenant-database-user')
+            && $this->userExists($connection, $config['username'])
+        ) {
```
ArlonAntonius: Can you return this to its original state?
I don't see any code changing except for some linting

src/Generators/Webserver/Database/Drivers/PostgreSQL.php
mikebronner: In my use-case I have only one database user, that is the main database user used on the System connection as well as all other Tenant connections. Removing privileges for a user would remove all the items they owned, regardless of schema they were in. That is dangerous, and in my case broke functionality.

src/Generators/Webserver/Database/Drivers/PostgreSQL.php
mikebronner: I had to make this change to get the checks to pass.

src/Generators/Webserver/Database/Drivers/PostgresSchema.php
mikebronner: My thoughts on this method is that we shouldn't be executing this code if our intent is not to drop the schema. And if our intent is to do so, why wouldn't we follow through with it? The problem I ran into when not adding the CASCADE is that the schema would never be dropped automatically, because the tables it contained were never dropped, as I disabled blanked removal of privileges (see change above), which is very risky.

PR #970, opened 2020-12-29
Drop AbstractModel

nzulfigarov:
Hi everyone. In this PR I:

1. Wiped out AbstractModel class.
2. Changed AbstactModel to base laravel Eloquent Model in Validator.

Reasons:
1. [I got the same Issue](https://github.com/tenancy/multi-tenant/issues/790), I can't interact with WebsiteRepository because Validator for some not understandable reasons want to see in arguments of its methods only AbstractModel, but how that is possible when I override Website & Hostname models and I want to inherit from my own base model?
2. AbstractModel is completely empty. No code is used at all, not a single line.
3. Validator uses only Eloquent fields/methods. 
4. If contributors in future want to use some reference on specific methods/fields, we do not need to create a whole abstract class for this kind of reasons, we need just define an Interface and define something we need in **Interface**. If we develop this package in Interfaces way, everyone like me who wants to create its own Website and Hostname models - can just implement a few interfaces, no abstract models needed.
Remember about **[Interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle)** 

P.S Sorry for misspelling/grammar mistakes - English is not my native language.


nzulfigarov:
Is anyone still maintain this package?

ArlonAntonius:
Hi there,

We are, we just have been really busy with tenancy/tenancy.

Some tests are failing on this PR, could you take a look at that? Those are some really serious issues by the look of it.

@bkintanar Can you make sure this PR comes through?

nzulfigarov:
Thank you for the response, I had almost given up hope...  :)

In tests in CircleCI fail only **php-7.3-apache** and **php-7.4-apache** with message only on PGSQL:
```
Caused by
PDOException: SQLSTATE[08006] [7] FATAL:  password authentication failed for user "6900748c752b460a927861e1d04943d5"
```
Nginx all good, on pg & mysql both. 
I think the error is caused by incorrect configurations in CI/CD, because my changes hardly affect anything. 

@bkintanar @ArlonAntonius

nzulfigarov:
All tests passed.

PR #979, opened 2021-01-30
Possibility to set middleware to tenant routes

dneykov:
This PR add possibility to set middlewares to tenant specific routes.

Reason
- In my tenant specific views when I try to access global variable `$errors` it's undefined. That is because dynamic loaded `routes.php` file is missing `web` group middleware. Applying the changes from that PR that problem is solved.

### DIFF
--- a/assets/configs/tenancy.php
+++ b/assets/configs/tenancy.php
@@ -355,6 +355,12 @@ return [
              * Prefix all tenant routes.
              */
             'prefix' => null,
+            
+            /**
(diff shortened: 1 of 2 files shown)


fletch3555:
Can't you just manually add the middleware with a route group in the tenant routes file? We also don't add a controller prefix (before laravel 8 at least) or any of the other stuff that routes/web.php gets.

I'm not specifically opposed to this, but I'm not sure it's really adding any benefit since it's already possible to do what you're asking.

PR #996, opened 2021-04-30
Fix HostnameActions aborting request to the Sytem site

duckzland:
Adding a check to check if the request is for the System by matching the http host against user configured default hostname and only abort if it doesn't match

### DIFF
--- a/src/Middleware/HostnameActions.php
+++ b/src/Middleware/HostnameActions.php
@@ -68,8 +68,12 @@ class HostnameActions
             if (!$request->secure() && $hostname->force_https) {
                 return $this->secure($hostname, $request);
             }
-        } else {
-            $this->abort($request);
+        } 
+        
+        else {
+            if ($request->getHttpHost() !== config('tenancy.hostname.default')) {
+                $this->abort($request);
+            }
         }
 
         return $next($request);


fletch3555:
I'm not sure I understand the intent here.  This code is aborting if a `Hostname` record is found (i.e. associated with a tenant), AND that tenant has `force_https` set to true, AND the incoming request is not secure (i.e. not using HTTPS).  It seems reasonable to abort in that case.  What does a system hostname have to do with this case?  HostnameActions middleware shouldn't apply to non-tenant requests

duckzland:
Well, the HostnameActions is loaded via the default tenancy.php config under the middleware section [ at least under version 5.6.2 ], so not sure if that is intended for tenants only or including system site.

In my case, the system site has route that uses subdomain like api.system.com, which is a valid subdomain for system callback but not a valid tenant, thus need the fix to check for fqdn and pass it to system and let system decide if it is a valid callback or not.

fletch3555:
So do you have a record in `websites` AND `hostnames` for api.system.com?

duckzland:
no, there is no record at websites for api.system.com as it is not a "tenants" but a sub domain of "system"

fletch3555:
That's why I was asking.  If you don't have a "website" record for it, then it won't be loaded as a tenant and therefore this code shouldn't run.  I think you might have something else going on.  Can you ask in our discord server so we can have a more direct discussion on this?

duckzland:
It wasn't intended to be run as a "tenant", but it should be run as a "system/landlord" but multitenancy will treat it as a tenant thus the middleware is fired.

Maybe I'm missing something? can you show me the code where the middleware is locked to tenants request only?

PR #1001, opened 2021-05-26
Trim slashes from request uri before redirecting

JaZo:


ArlonAntonius:
Hi @JaZo,

Could you specify what problems you are bumping into in this case? This looks like an adjustment for a really specific situation and looks like something that might be adjustable with some server configuration.

sniper7kills (approved):
Approved after internal review.

bkintanar:
Approved after internal review.

JaZo:
Thanks for the quick response!

PR #1028, opened 2022-10-17
Remove phpseclib from composer.json because of conflict with league/filesystem-sftp-v3

markieo1:
Hello,

In this project phpseclib is listed as a dependency. This library isn't used in the whole project.
Also this causes issues with the https://github.com/thephpleague/flysystem-sftp-v3 which requires phpseclib ^3.0.

Thanks in advance!


markieo1:
Tests are currently failing because of issues with the database in CI. I need some help with that.

markieo1:
Anyone any idea why the tests are failing?

bkintanar:
@markieo1 some pgsql permissions. just leave it for now.

PR #1036, opened 2023-02-20
[5.x] Laravel 10 Support

ArlonAntonius:


bkintanar (approved):
LGTM.

jonquintero:
Hi guys  how are you.
Any  new change about this update?
Thanks

PouleR:
Are you planning to merge this PR so we can update to Laravel 10? Thank you in advance!

PouleR:
@ArlonAntonius please note that the usage of the $dates property has been removed.
This is still being used in the Hostname model: `protected $dates = ['under_maintenance_since'];` and should be rewritten to the casts property:

src/Models/Hostname.php:
```
protected $casts = [
    'under_maintenance_since' => 'datetime',
];
```

mattvb91:
any update on this?

PouleR:
@ArlonAntonius I've create a PR for you to update this pull request, this should fix the build so we can hopefully merge this.

jonquintero:
Hi @PouleR Can you make this works with php 8.2 too ? Thanks

PouleR:
> Hi @PouleR Can you make this works with php 8.2 too ? Thanks

@jonquintero I just added php-8.2 to the circle ci jobs!

jonquintero:
Hi guys, how are you? I want to ask if there is any news about this ?

PouleR:
@ArlonAntonius how about merging this PR? Would be really appreciated!

ArlonAntonius:
You're a hero @PouleR, thanks for all the effort 🚀

mattvb91:
Thanks @PouleR. Could we also get a tag so we can pull it?

tvbeek:
@PouleR @ArlonAntonius I'm not sure if you think about this, but this creates a breaking change in a minor release. That is because you drop Laravel 9 support.

Is that intended? Or does there need to be a new PR that bring back the support for Laravel 9.

ArlonAntonius:
@tvbeek You are completely right, that is a user error and it should be made backward-compatIble. I should have some time this weekend to fix that.
Only thing I am worried about is the PHP versions.

jonquintero:
Is not possible to put a tag for laravel 10 support for example 5.9.1 while for laravel 9 you would use something like 5.9.0 ?

tvbeek:
@jonquintero I'm interested how you see that working for new tags? A 5.10.1 for Laravel 10 and 5.10.0 for Laravel 9? And besides that it is still a breaking change 😃

ArlonAntonius:
@tvbeek is right, it is a breaking change and should not have been. I have a PR ready in order to add back L9 compatibility. If you guys have ideas, let's move it there :) #1041

PR #1041, opened 2023-08-12
[5.x] Add Laravel 9 Support back to 5.X

ArlonAntonius:
Might still need CircleCI changes


tvbeek:
Thanks for the fix 👍

ArlonAntonius:
@tvbeek thanks for pointing it out 😉

PR #1047, opened 2024-05-11
Upgrade to Laravel 11.

mikebronner:


ArlonAntonius:
@mikebronner Thank you for starting the effort on this, let us know if there's anything we can do to help update this package.

chintan13:
Looking for upgrade to laravel 11, thank you!

mikebronner:
After testing this for several months now on my clients I have not encountered any issues. @ArlonAntonius marking this ready for review. Although I have not used the two file methods. Could use some help implementing those.

ArlonAntonius:
@mikebronner have you really been tesing it in production like this?
So simply CI is broken?

mikebronner:
@ArlonAntonius  Yes, correct, but with the caveat that we have not been using file uploads in our multi-tenant apps. I assume the CI issues are from Laravel 11 now requiring PHP 8.2+, but I am unable to see why the CI is failing (when I click on the link it asks me to log in).

src/Website/Directory.php
```diff
@@ -102,6 +102,16 @@ public function path(string $path = null, $local = false): string
...
 
+    public function putFile($path, $file = null, $options = [])
+    {
+        // tba
+    }
+
+    public function putFileAs($path, $file, $name = null, $options = [])
+    {
+        // tba
+    }
```
ArlonAntonius, line 113: @mikebronner Maybe the only thing that still needs to be added are these methods to make sure nobody's application actually fails. Other than that: I'm fine with merging this

ArlonAntonius (changes requested):

